PR #1, opened 2016-02-24
Remove the v1 and unversioned routes

rowanmanning:
v1 and unversioned routes now redirect to their v2 counterpart. If a static bundle exists for the given route, then it is served from the file system.

The original CSS bundler has been removed and the OBT CSS bundler has been renamed to replace it.


lucas42:
When this goes live, please can you make sure Ops are aware of the change and its potential impact? Also, when deploying, post in the #ft-tech-incidents channel on slack so that other teams know it's happening.  Thanks.

alicebartlett:
@lucas42 When you say "make sure ops are aware" what do you mean by that? Is there a slack channel we should be using?

PR #6, opened 2016-02-25
Put together a suggested testing structure

rowanmanning:
I've put together this as a suggested new structure for testing this project. Completely open to suggestions – wanted to get something up to start a discussion. Feedback would be much appreciated! I'll justify a few decisions here too.

The key files to look at are:
- [Example unit test](https://github.com/Financial-Times/origami-build-service/pull/6/files#diff-e5fee1c925d25a3fbc5c1399819bdc4e)
- [Example integration test](https://github.com/Financial-Times/origami-build-service/pull/6/files#diff-e090eea4da438d3eb1e8fc7a1c91ddc9)

### Directories

Splitting unit and integration tests makes it easier to find what you're looking for as well as enabling them to be run separately. Considering how slow the integration tests can be, this makes unit testing a lot less painful. Also coverage reports aren't that useful with integration tests, and they may give false positives. Also the unit test directory structure mirrors the root of the project, this aids in finding the tests for individual modules.

### Libraries

I'm using Mocha, Sinon, Mockery, and Chai. We're already using Mocha and Chai, and Sinon is being used in a lot of other projects. Mockery is a new addition and allows you to mock `require`.

### Mocking

The unit tests make heavy use of mocking, which speeds them up like crazy. Also coverage reporting is far more accurate/useful. There's overhead to this approach, but I think it's worth it for the increased speed. I realise the test I used as an example isn't the clearest in retrospect. Happy to put together another test if you want a better demo.

### Build tools

Currently I've added `test-unit` and `test-integration` npm scripts. This makes it easy to run them separately and at some point `npm test` would just alias these, something like `npm run test-unit && npm run test-integration`. I'm a big fan of Make, and may suggest moving to use that once Next's investigation into it is done.

### Going forward

If we end up changing the testing structure, it'll need to coexist with the current tests for a little while. This is made a little easier with the new directories, as there's no confusion about which tests are old vs new. I'd also document this in the README.


rowanmanning:
@AlbertoElias, @alicebartlett, @onishiweb, fancy taking a look?

onishiweb:
All the justifications for the approach are sound, I like the directory sorting and it's advantages. Mocking seems to make sense, would be interested in knowing what the steps would be to do these tests without it just as an example.

Unsure about the switch to `make`, what advantages would this have over sticking with running `npm test` aliasing the new testing scripts?

rowanmanning:
Re Make there aren't _loads_ of advantages over npm. The main one I can see is easily stringing tasks together that are dependent on each other. E.g.

```
test: lint test-unit test-integration
```

I think mostly though it's personal preference, which isn't much of an argument :)

AlbertoElias:
I really like this, very easy to follow. Brilliant work :) Only thing is how to disable the logs cleanly and like we spoke in person, I think npm scripts are fine for these very simple use cases. Next's make scripts are very long and complex, and they have many.

PR #10, opened 2016-03-01
Replace the shrinkwrap URL with a deprecation notice

rowanmanning:


onishiweb:
:+1: apart from the failing tests

AlbertoElias:
Having am update guide is a bit complicated as each bundle can have different issues. I'd rather point them to our email

rowanmanning:
I tweaked the static bundle notice to include an email rather than the URL. Does this make more sense? 25ba1464f4f0fd3e63f2a31d000d3c5367416a72

PR #16, opened 2016-03-02
Add missing unit tests

rowanmanning:


test/unit/lib/output.test.js
```diff
@@ -0,0 +1,85 @@
...
+		});
+
+		it('should have a `stream` property containing a stream representation of `bufferedContent`', function() {
+			assert.notCalled(streamCat);
+			assert.strictEqual(instance.stream, streamCat.firstCall.returnValue);
+			assert.calledOnce(streamCat);
+			assert.deepEqual(streamCat.firstCall.args[0], [bufferedContent]);
+		});
+
+		it('should have a `pipe` method', function() {
```
AlbertoElias: Is this one and the following necessary? They seem to be more streamcat/node.js tests

rowanmanning: I'd agree if we were extending Stream or something, but the Output object implements the pipe method itself https://github.com/Financial-Times/origami-build-service/blob/master/lib/output.js#L25

test/unit/lib/utils/uniqueid.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+		describe('when `id.length` is greater than `maxLength`', function() {
+			let id;
+
+			beforeEach(function() {
+				crypto.createHash.returns({
+					digest: sinon.stub().returns('foobarbazqux1234567890abcdef'),
+					update: sinon.stub()
+				});
+				id = 'hello-world-how-are-you-today-foo-bar-baz';
```
AlbertoElias: I'm doing good thanks, I've got Jupiter behind me so I'm trying to cast a shadow on it, wish me luck!

AlbertoElias:
Looks good!

PR #17, opened 2016-03-03
Static bundle refactor and tests

rowanmanning:


lib/static-bundle.js
```diff
@@ -0,0 +1,30 @@
...
+
+const fs = require('fs');
+const path = require('path');
+const pfs = require('q-io/fs');
+const uniqueid = require('./utils/uniqueid');
+const URL = require('url');
+
+const defaultStaticBundlesDirectory = path.resolve(__dirname, '../deprecated/static-bundles');
+
+function getStaticBundleStream(url, staticBundlesDirectory) {
```
AlbertoElias: If this has now been moved out, would it make sense to put the file in the `deprecated` directory so it's easier to remove sometime in the future and also doesn't divert attention from the main functionality?

rowanmanning: Good idea, I'll move it :)

test/unit/mock/express.mock.js
```diff
@@ -4,5 +4,16 @@ const sinon = require('sinon');
 
 module.exports = sinon.stub().returns({
 	disable: sinon.spy(),
-	use: sinon.spy()
+	use: sinon.spy(),
```
AlbertoElias: I thought lingering commas weren't proper JSON

rowanmanning: Oops! Linter should probably pick that up

AlbertoElias: Yes, it should. Don't know what might have gone wrong =/

AlbertoElias:
Looks good :+1:

Issue #18, opened 2016-03-09
Suggestion: Use `.env` for development environment variables instead of `docker-compose.yml`


JakeChampion:
Currently the README advocates [storing confidential environment variables inside `docker-compose.yml` for local development](https://github.com/Financial-Times/origami-build-service/tree/docker#configuration). This becomes a bit messy because that file is included within git but will become personal to each developer's set-up (API keys, auth tokens, database names etc). A developer may accidentally commit their environment variables into the git history (which is a bad thing _tm_).

Have you thought about using [`dotenv`](https://www.npmjs.com/package/dotenv) instead of placing the environment variables within `docker-compose.yml`?

`docker-compose.yml` is commited into the repo whereas `.env` can be ignored in `.gitignore`. This means confidential env vars can be stored in it safely and because the file is ignored by git it won't be deployed to production.

onishiweb:
:+1:

Issue #7, opened 2016-02-25
Add Continuous deployment


AlbertoElias:
Easy to do with circle

rowanmanning:
Update on this: we can't quite add continuous deployment because of this issue https://github.com/heroku/heroku-docker/issues/138

We have two options:

1. Wait for this issue to be fixed, which could take a while.
2. Move to official Heroku Docker support, which is currently in private beta.

JakeChampion:
I don't think the issue will get fixed as Heroku are no longer advocating use of their Heroku Docker plugin which creates slugs. This means the option left is to use their Docker registry which is in private beta.

rowanmanning:
I think we've decided that 2 is the better option because of this :)

rowanmanning:
We're doing this now

PR #19, opened 2016-03-09
Docker Update and Continuous Deployment

rowanmanning:
### Outline

This PR contains the work required to get the build service continuously deploying, addressing #7. In order to do this we've had to overhaul the way that Docker is used, moving to Heroku's improved Docker support (which is currently in private beta). Here's what's been done:

- **Improved Docker support:** Deployment is now done by pushing a built Docker image to Heroku's registry. I've simplified the commands required to do this:
  - Local development: `docker-compose build` and `docker-compose up`
  - Deploy: `npm run build` and `npm run deploy`

- **Continuous deployment:** The above build/deploy commands no longer need to be run manually as they'll be run in CircleCI. Any push to the `master` branch will be deployed to QA once the tests have passed. Production is not deployed to through CI (see pipelines).

- **Pipelines:** We're using a Heroku pipeline to deploy from QA to production. This makes a production deploy a simple button press, and the image that gets pushed into production is the exact same code that is running successfully in QA.

- **Cleanup:** Removal of any files which are no longer necessary, and a cleanup of documentation/environment variables which are no longer used.

### What Needs Doing Before/After Merge

First we need to agree that this is approach to continuous deployment that we'd like to take. I can demo everything to the team and we can have a chat about it.

Code-wise, everything is currently deploying happily to `origami-buildservice-rowan-qa` and `origami-buildservice-rowan`. We need to do a few things before this is ready to merge:

1. In [`circle.yml`](https://github.com/Financial-Times/origami-build-service/blob/docker/circle.yml#L17), change the deployment branch to `master`.
2. Change references in [`package.json`](https://github.com/Financial-Times/origami-build-service/blob/docker/package.json#L29-L30) from `origami-buildservice-rowan-qa` to `origami-buildservice-qa`.
3. Merge this PR (maybe squashing the commits first).
4. Tag a new version on master and push it (wait for the deploy to work).
5. Verify that the new version is on QA and everything works.
6. Create a Heroku pipeline from `origami-buildservice-qa` to `origami-buildservice-eu`.
7. Promote the QA deploy to production.
8. Verify that the new version is on production and everything works.

### Next Steps

These are things that I'd like to get done after this has been merged, but don't need to be done to get continuous deployment up-and-running:

- #18: using `.env` instead of `docker-compose.yml` for environment variables
- #21: run the integration tests against a running container in CI


rowanmanning:
@JakeChampion, @AlbertoElias, @alicebartlett, @onishiweb, @j-mes, would be good to get your thoughts/feedback on this. We can discuss in-person as a group if it's easier.

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+indent_style = tab
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[**{.js,.scss,.html,.mustache,.md,.yml}]
```
JakeChampion: `.yml` forbids tabs

rowanmanning: Oops, I got that wrong, and preventing yml from having tabs is the entire reason I added this file :joy:

Dockerfile
```diff
@@ -1,44 +1,22 @@
-FROM heroku/cedar:14
+FROM mhart/alpine-node:4
```
JakeChampion: It is not mentioned anywhere in this PR why the change from cedar to alpine was made.

Below is the explanation given in the [heroku-docker-node-alpine-example](https://github.com/Financial-Times/heroku-docker-node-alpine-example) repository

>Heroku's Cedar base image is based upon Ubuntu, which was not developed with the idea of being run inside a container, incidentally making the image size rather large. Heroku's Cedar also installs a lot of packages to cover the needs of most Heroku developers' applications.

>Alpine Linux is built around musl libc and busybox. This makes it smaller and more resource efficient than traditional GNU/Linux distributions. An Alpine Linux image requires no more than 8 MB.

>A smaller base image with fewer dependencies gives benefits such as a faster build/deploy time, smaller artifacts created for each build, and a smaller surface area to check for security vulnerabilities.

Dockerfile
```diff
@@ -1,44 +1,22 @@
...
-ENV HOME /app
+# Set the working directory
 WORKDIR /app
-RUN mkdir -p /app/.profile.d
 
-# Node will be installed in /app/heroku/node, app in /app/user
-RUN mkdir -p /app/heroku/node
-RUN mkdir -p /app/user
+# Install additional dependencies required to build modules
+# TODO comment on _why_ these are here
```
JakeChampion: I believe these are here because of `node-sass` and possibly because of `nodemon` and `webpack`?

Dockerfile
```diff
@@ -1,44 +1,22 @@
...
-ENV PATH /app/heroku/node/bin:/app/bin:/app/user/node_modules/.bin:$PATH
-RUN echo "export PATH=\"/app/heroku/node/bin:/app/bin:/app/user/node_modules/.bin:\$PATH\"" > /app/.profile.d/nodejs.sh
+# Install Node.js dependencies
+COPY package.json /app/
+RUN npm install -g nodemon
+RUN npm install --production
+RUN npm cache clean
 
-# Configure git to use HTTPS instead of SSH
-RUN echo '[url "https://"]\n    insteadOf = git://' > /app/.gitconfig
```
JakeChampion: This may still be required for the build-service to pull bower components from git correctly.

rowanmanning: Oh yeah, good point. On it

README.md
```diff
@@ -2,81 +2,88 @@
...
+brew cask install dockertoolbox docker-compose
+```
 
-You may now have to change the owner of your `.docker` directory if it is owned by root:
+Create a virtual machine to run the application's containers, and put that machine's config into your environment. The default size didn't appear to be large enough so this will create one with an increased disk size:
 
-    chown -R `whoami` ~/.docker
+```sh
+# Create a Docker machine
+docker-machine create --driver virtualbox --virtualbox-disk-size "50000"
```
JakeChampion: That's a big disk size!

README.md
```diff
@@ -2,81 +2,88 @@
...
-    echo "eval $(docker-machine env dev)" >> ~/.profile
+# Add the machine's config to your current environment
+eval $(docker-machine env)
+```
 
 Find out the IP address of the machine:
 
-    docker-machine ip dev
+```sh
+docker-machine ip dev
```
JakeChampion: remove `dev`

README.md
```diff
@@ -2,81 +2,88 @@
...
+docker-compose build
+docker-compose up
+```
 
-Now you can access the app at the IP address discovered earlier, over HTTP on port 8080.  So eg in the browser go to http://192.168.99.102:8080/ (if that was the IP given by `docker-machine ip dev`)
+Now you can access the app at the IP address discovered earlier, over HTTP on port 8080:
 
-To SSH into the web container, you first need to SSH into the Docker VM, and then into the container you want:
+```sh
+open "http://$(docker-machine ip dev):8080/"
```
JakeChampion: remove `dev`

README.md
```diff
@@ -2,81 +2,88 @@
...
-* `VIEW_CACHE_PATH`: Path on disk to use to cache view templates in Twig
-* `DEBUG_KEY`: String, if set in a `Debug` HTTP header, will set dev mode to true for that request only.
-* `GITHUB_USERNAME`: A GitHub username which has access to any private repositories that bower needs to see
-* `GITHUB_PASSWORD`: The GitHub password corresponding to `GITHUB_USERNAME`
+* `PORT`: The port to run the application on. This is set by docker-compose locally, and Heroku in production.
+* `NODE_ENV`: Standard Node convention for specifying which type of environment we're in, 'development' or 'production'.
+* `SENTRY_DSN`: URL of the Sentry project to use to collect runtime errors, exceptions and log messages.
+* `GRAPHITE_HOST`: The host of the Graphite server used to gather metrics.
+* `LOG_LEVEL`: Syslog-compatible level at which to emit log events to stdout ('trace', 'debug', 'info', 'warn', 'error', or 'crit').
+* `GITHUB_USERNAME`: A GitHub username which has access to any private repositories that bower needs to see.
```
JakeChampion: For future work, perhaps switching to an OAuth token may be a safer approach?

circle.yml
```diff
@@ -9,3 +11,12 @@ test:
     - npm run test-old
     - npm run test-unit
     - npm run test-integration
+
+deployment:
+  production:
+    branch: docker
+    commands:
+      - 'echo "machine api.heroku.com login rowan.manning@ft.com password ${HEROKU_AUTH_TOKEN}" >> ${HOME}/.netrc; chmod 0600 /home/ubuntu/.netrc;'
```
JakeChampion: Note: This application will fail to deploy from CI is @rowanmanning ever renews his Heroku authentication token.

README.md
rowanmanning: Good idea. Open an issue?

docker-compose.yml
```diff
@@ -1,18 +1,13 @@
...
+  volumes:
+    - "./lib:/app/lib"
+    - "./deprecated:/app/deprecated"
+  ports:
+    - "8080:8080"
+  environment:
+    PORT: 8080
+    NODE_ENV: "development"
+    LOG_LEVEL: "trace"
+  command: "nodemon --legacy-watch index.js | ./node_modules/.bin/bunyan -o short"
```
JakeChampion: Ah, this is why in the Dockerfile `nodemon` is installed globally, I was struggling to think of a reason for this. Makes sense now.

package.json
```diff
@@ -60,7 +62,6 @@
     "chai": "^3.5.0",
     "eslint": "^2.2.0",
     "istanbul": "^0.4.2",
-    "mkdirp": "^0.5.1",
     "mocha": "^2.4.5",
     "mockery": "^1.4.0",
     "nodemon": "^1.9.1",
```
JakeChampion: Nodemon can be safely removed from here.

circle.yml
rowanmanning: We could maybe do with an Origami Heroku account for this kind of thing? It would also fail if I ever get removed from the Heroku org.

README.md
JakeChampion: https://github.com/Financial-Times/origami-build-service/issues/20

alicebartlett:
this is a huge improvement to our deployment process, and this PR explains your changes really well too.

Dockerfile
rowanmanning: I added a description, missed that "TODO"

README.md
rowanmanning: Done

README.md
rowanmanning: Done

package.json
rowanmanning: I've removed it now

circle.yml
JakeChampion: I believe this is not something the Heroku Working Group within the FT advocate people to do.

rowanmanning: OK, we'll just need to keep an eye on it then

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+insert_final_newline = true
+charset = utf-8
+
+[**{.js,.scss,.html,.mustache,.md}]
+indent_style = tab
+trim_trailing_whitespace = true
+
+[**{.yml}]
+indent_style = space
+indent_size = 2
```
AlbertoElias: should me move this stuff into obt?

Dockerfile
```diff
@@ -1,44 +1,25 @@
...
-COPY docs /app/user/docs
-COPY test /app/user/test
-COPY tools /app/user/tools
-COPY deprecated /app/user/deprecated
+# Heroku ignores this command and will use their designated port set as an environment variable
+EXPOSE 8080
 
-# For dev, allow mounting of filesystem from outside the container
-VOLUME /app/user/lib
-VOLUME /app/user/node_modules/origami-build-tools
```
AlbertoElias: why remove this?

JakeChampion: It's not needed for production, only for development and is specified in `docker-compose.yml`

docker-compose.yml
matthew-andrews: you might be able to avoid this if you put `node_modules/.bin` on the $PATH, or by putting this command in an npm script and running that here:- `npm run dev`

**package.json:-**
```js
…
  "scripts": { "dev": "nodemon --legacy-watch index.js | ./node_modules/.bin/bunyan -o short" }
…
```

circle.yml
AlbertoElias: Yeah, this kind of sucks. I'm using my account for browser tests with nightwatch

docker-compose.yml
```diff
@@ -1,18 +1,13 @@
...
-    ports:
-        - "8080:8080"
-    environment:
-        PORT: 8080
-        NODE_ENV: "development"
-        LOG_LEVEL: "trace"
+  build: .
+  volumes:
+    - "./lib:/app/lib"
+    - "./deprecated:/app/deprecated"
```
AlbertoElias: I thought these still require the volumes being set in the Docker file

circle.yml
matthew-andrews: in Next we use my `HEROKU_AUTH_TOKEN` and push it out to all our CircleCI configs.  I can setup origami-build-service with this if you like and then we can manage this for you ^^

Dockerfile
rowanmanning: Yep, also Heroku's Docker support [explicitly doesn't allow volume mounting](https://devcenter.heroku.com/articles/docker-runtime-and-registry#runtime)

docker-compose.yml
rowanmanning: It works locally for me, the volumes mount and nodemon sees the file changes

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.e*
+.git
+.gitignore
+circle.yml
+coverage
+docker-compose.yml
+Dockerfile
+node_modules
```
triblondon: node_modules should presumably not exist outside the container anyway, right? But I guess this is worth having just in case.

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.e*
```
triblondon: bit cryptic. Is this meant to capture .env?

.eslintignore
```diff
@@ -1,3 +1,3 @@
 coverage
-test/testmodules
 test/integration/mock-modules
+test/testmodules
```
triblondon: Unnecessary change?

.dockerignore
rowanmanning: No, all the `.eslintrc` files etc. You're right, it's cryptic. I'll add the file names explicitly

Dockerfile
triblondon: Is alpine likely to continue to be maintained?  Is that a concern?

.dockerignore
rowanmanning: It _shouldn't_, but if someone runs npm install locally it's useful to ensure that it doesn't overwrite the modules installed in the container

Dockerfile
```diff
@@ -1,44 +1,25 @@
...
-RUN mkdir -p /app/user
-
-# Install node and configure PATH
-ENV NODE_ENGINE 4.3.1
-RUN curl -s https://s3pository.heroku.com/node/v$NODE_ENGINE/node-v$NODE_ENGINE-linux-x64.tar.gz | tar --strip-components=1 -xz -C /app/heroku/node
-ENV PATH /app/heroku/node/bin:/app/bin:/app/user/node_modules/.bin:$PATH
-RUN echo "export PATH=\"/app/heroku/node/bin:/app/bin:/app/user/node_modules/.bin:\$PATH\"" > /app/.profile.d/nodejs.sh
+# Install additional dependencies required to build modules
+# We need these to build native dependencies (mostly node-sass)
+RUN apk add --no-cache --update g++ gcc git make python
```
triblondon: Useful comment here :+1:

.eslintignore
rowanmanning: Yes, it is. My OCD wanted this to be in alphabetical order. I can revert if you like :)

README.md
```diff
@@ -2,81 +2,88 @@
...
 
 
-## Testing and monitoring
+## Testing
 
-The tests are split into unit tests, integration tests, and an older suite of tests that we're in the process of migrating. Use the following commands to run tests:
+The tests are split into unit tests, integration tests, and an older suite of tests that we're in the process of migrating. To run tests on your machine you'll need to install [Node.js](https://nodejs.org/) and run `npm install`. Then you can run the following commands:
 
 ```sh
 npm test                  # run all of the tests
```
triblondon: why are we not running tests in the container?

circle.yml
triblondon: For the unitiated, what does the `machine` command do and how is it available in the circle environment?

docker-compose.yml
triblondon: Agreed - doesn't this break watch?  So you'd need to manually rebuild and restart the container for every source code change?

Dockerfile
```diff
@@ -1,44 +1,25 @@
...
-COPY app.json /app/user/
-COPY about.json /app/user/
-COPY index.js /app/user/
-COPY lib /app/user/lib
-COPY docs /app/user/docs
-COPY test /app/user/test
-COPY tools /app/user/tools
-COPY deprecated /app/user/deprecated
+# Heroku ignores this command and will use their designated port set as an environment variable
+EXPOSE 8080
```
triblondon: We also configure ports in docker-compose for dev, so do we need an expose in Dockerfile at all?

README.md
rowanmanning: I want to, but it requires a little refactoring and this PR is already pretty big. I've opened #21 to remind me to fix this.

triblondon:
Very nice.

docker-compose.yml
rowanmanning: Nope, watch still works for me. I can investigate further to verify but my changes appear on the container.

Dockerfile
rowanmanning: Possibly not, I'll try removing it.

circle.yml
matthew-andrews: @triblondon run this on your mac `cat ~/.netrc`

Dockerfile
JakeChampion: Not a major concern, the community around Alpine is large.

circle.yml
JakeChampion: You can create an Origami heroku account by the way, I spoke with the Heroku Working Group and they verified this is a good approach for Origami to take with it's deployments.

docker-compose.yml
JakeChampion: Yes, I think nodemon shouldn't be globally installed in the dockerfile and should instead a dev-dep and executed the way @matthew-andrews mentioned above.

.dockerignore
JakeChampion: :+1:

Dockerfile
rowanmanning: Actually I think we want to leave this in, so that in a non-Heroku environment (like CI) we don't need to specify a port. We want to run the production container on CI for integration tests.

docker-compose.yml
matthew-andrews: If you do that you'll still need `npm install nodemon` because the `npm install --production` will stop nodemon from being installed

.editorconfig
rowanmanning: We could have a copy in there, yes, but the editorconfig plugins rely on there being a `.editorconfig` file in the project root.

docker-compose.yml
rowanmanning: Yep, I've investigated further and `nodemon` still picks up file changes.

.editorconfig
AlbertoElias: Do you mean the IDE plugins? We should probably discuss what's the best way
to use editorconfig

-- 

------------------------------

*This email was sent by a company owned by Financial Times Group Limited 
("FT Group <http://aboutus.ft.com/corporate-information/#axzz3rajCSIAt>"), 
registered office at Number One Southwark Bridge, London SE1 9HL.  
Registered in England and Wales with company number 879531. This e-mail may 
contain confidential information. If you are not the intended recipient, 
please notify the sender immediately, delete all copies and do not 
distribute it further.  It could also contain personal views which are not 
necessarily those of the FT Group.  We may monitor outgoing or 
incoming emails as permitted by law.*

package.json
```diff
@@ -24,9 +25,9 @@
     "test-instrument": "NODE_ENV=test ./node_modules/.bin/istanbul cover ./node_modules/mocha/bin/_mocha -- ./test/unit --recursive",
     "test-coverage": "npm run test-instrument && ./node_modules/.bin/istanbul check-coverage --statement 90 --branch 90 --function 90",
     "lint": "./node_modules/.bin/eslint .",
-    "deploy-prod": "git describe --tags > ./appversion; heroku docker:release --app origami-buildservice-eu; rm -f ./appversion",
-    "deploy-qa": "git describe --tags > ./appversion; heroku docker:release --app origami-buildservice-qa; rm -f ./appversion",
-    "build": "touch ./appversion; docker-compose build; rm -f ./appversion"
+
+    "build": "docker build -t registry.heroku.com/origami-buildservice-qa/web .",
+    "deploy": "docker push registry.heroku.com/origami-buildservice-qa/web"
```
triblondon: Just been thinking a bit more about this.  It seems like these scripts are a mixture of tasks run inside the container and tasks run outside the container.  Maybe we should follow Next and use Make to standardise outside-container-tasks like building the container and deploying it?

JakeChampion: Sure thing, these dont depend on node modules so I see no issue extracting into a makefile.

rowanmanning: I'm definitely up for adding a Makefile

Issue #22, opened 2016-03-15
Find a replacement for the diskspace module


rowanmanning:
The module is unreliable and it gets free space values by parsing a table output by the `df` command. The `df` command output is different on different Linux distributions, which caused some issues when moving to Alpine.

Also we currently maintain a forked version of the module and so have to install it through git. We could remove git as a dependency and reduce the container size if we find a good npm-hosted module.

JakeChampion:
Potential alternative which uses statvfs for *nix systems -- https://www.npmjs.com/package/diskusage

JakeChampion:
Heroku creates a [fresh filesystem whenever a dyno is restarted](https://devcenter.heroku.com/articles/dynos#ephemeral-filesystem). Heroku also [restart dynos every 24 hours (+- 216 minutes)](https://devcenter.heroku.com/articles/dynos#restarting), which means the filesystem is cleared once a day. Do we know if the filesystem use has been an issue before?

rowanmanning:
I don't think this is happening, at least not any time soon. Gonna close this for now, we don't even use Alpine any more which is why we were originally gonna do this

PR #23, opened 2016-03-15
Use our Alpine-friendly version of diskspace

rowanmanning:
I've had to push an Alpine-specific version of the diskspace module, because `df` on Alpine has different output. This caused the inodes check on the `__health` endpoint to fail. This highlights how fragile the diskspace module is, I've opened an issue to find a replacement: #22.

Here's the diff for the `alpine` branch of diskspace: https://github.com/Financial-Times/diskspace.js/compare/master...Financial-Times:alpine

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "bunyan": "^1.5.1",
     "cheerio": "^0.17.0",
     "commander": "^2.3.0",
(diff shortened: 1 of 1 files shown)


JakeChampion:
LGTM :+1:

PR #25, opened 2016-03-18
Replace use of string.contains

rowanmanning:
Not sure how/why they ever made it in – they should never have worked.


AlbertoElias:
Didn't it use to be contains and then the api was renamed to include? Maybe a recent v8 update broke it?

AlbertoElias:
Looks good!

PR #41, opened 2016-04-05
Cache the built Docker image on CI

rowanmanning:
This drastically improves the speed of QA deployments with a primed cache.

Before: **~13 minutes**
After: **~6 minutes** (not 100% sure yet as haven't tested the push, seeing ~3 minutes with no push)


JakeChampion:
LGTM :+1:

PR #42, opened 2016-04-05
Implement automated change requests

rowanmanning:
This adds two new Make tasks:

```sh
make change-request-qa
make change-request-prod
```

Both of these use `tools/change-request.js`, which sets some defaults for the created change request. I based these defaults on what the Next team have been sending.

This tool uses the Node.js module [`konstructor-api-client`](https://github.com/Financial-Times/konstructor-api-client-node). We depend on an Alpha version while we're testing, we'll switch to a semver range once we release `1.0.0`.

## QA Deploy

Because CI is outside of our network we need a Konstructor API key, which is set using an environment variable on CircleCI.

The tool needs a GitHub username; in CI we can get this from the `CIRCLE_USERNAME` environment variable. We use this to find an associated `ft.com` email address via [this spreadsheet](https://docs.google.com/spreadsheets/d/1mbJQYJOgXAH2KfgKUM1Vgxq8FUIrahumb39wzsgStu0/edit#gid=0). Konstructor then uses the email address to set the appropriate change request owner in Salesforce.

## Production Deploy

When deploying to production through Make, we now also attempt to open a change request. Because we have no `CIRCLE_USERNAME` environment variable locally, we have to set our GitHub username. The documentation outlines this:

```sh
make promote
```

becomes:

```sh
GITHUB_USERNAME=yourgithubusername make promote
```

## Failure

If the change request fails, we don't fail the build or prevent a deploy. This is because the next team have observed some random failures in the API, and cancelling a deploy because of these is disruptive.

It's up to the person who's deploying to decide if it's important enough to try again, or manually log a change request through Salesforce. It's likely that we won't bother for QA, but will for production deploys.

### DIFF
--- a/tools/change-request.js
+++ b/tools/change-request.js
@@ -1,0 +1,72 @@
+#!/usr/bin/env node
+'use strict';
+
+const KonstructorApiClient = require('konstructor-api-client');
+const pkg = require('../package.json');
+const program = require('commander');
+
+// Command-line configuration
+program
+	.usage('[options]')
+	.option(
+		'-g, --gateway <gateway>',
+		'a gateway to log the change request to, one of "internal", "konstructor", "mashery", "test". Default: "test"',
+		'test'
+	)
+	.option(
+		'-a, --api-key <key>',
+		'an API key for use with "konstructor" or "mashery" gateways',
+		process.env.KONSTRUCTOR_API_KEY
+	)
+	.option(
+		'-e, --environment <environment>',
+		'the environment the change request applies to. One of "Production", "Test", "Development", "Disaster Recovery". Default: "Test"'
+	)
+	.parse(process.argv);
+
+// Build change request
+const type = 'releaselog';
+const username = process.env.CIRCLE_USERNAME || process.env.GITHUB_USERNAME;
+const summary = `Releasing Origami Build Service version ${pkg.version} to ${program.environment}`;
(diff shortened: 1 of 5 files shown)


tools/change-request.js
```diff
@@ -0,0 +1,72 @@
...
+const closeData = {
+	notifyChannel: notifyChannel
+};
+
+// Create a change request
+console.log('Creating a change request');
+const apiClient = new KonstructorApiClient(program.gateway, program.apiKey);
+KonstructorApiClient.getEmailFromUsername(username)
+	.then(email => {
+		openData.ownerEmailAddress = openData.resourceOne = closeData.closedByEmailAddress = email;
```
onishiweb: This is neat, I didn't realise this could be done to set multiple resources to the same value.

onishiweb:
LGTM :+1:

tools/change-request.js
rowanmanning: Good eh? Useful to know how/why it works: [JavaScript Operator Precedence - Associativity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Associativity)

alicebartlett:
As per usual, this is a great pull request, well explained, succinct. The change looks good too. I'm going to go right ahead and click that merge button.

PR #44, opened 2016-04-11
Catch EIA_AGAIN errors and output them as a 502

rowanmanning:
EAI_AGAIN means a temporary failure in name resolution, or the name
server returned a temporary failure indication. The solution is to try
again later.

I think this makes the most sense to expose to the user as a 502
"Bad Gateway" error. This will also now get logged to Graphite as a
remote IO error rather than bothering us in Sentry with an error
that we can't actually resolve.

For more information, see:
http://www.codingdefined.com/2015/06/nodejs-error-errno-eaiagain.html


onishiweb:
Yeh I think that makes sense. Not much we can do about the error so the 502 makes the most sense to me too. Looks good 👍

PR #45, opened 2016-04-11
Fix shrinkwrap URLs for Git sources

### LINKED ISSUE #26: Shrinkwrap URLs and non-semver versions
The way that shrinkwrap URLs are calculated doesn’t account for non-semver versions in the bower config (e.g. a git URL or GitHub username/repo).

For example: `o-techdocs` includes `highlight.js`, but from our shim version: https://github.com/Financial-Times/o-techdocs/blob/master/bower.json#L7

Currently when the shrinkwrap URL is generated it references `highlight.js@undefined` rather than `Financial-Times/highlight.js-shim@<commit-hash>` (which works).

I'm looking into whether I can fix this without breaking other shrinkwrap URLs. Unfortunately this means that current shrinkwraps will never work, unless I add in a specific case to always serve our version of `highlight.js` (which would be super hacky).

**Steps to reproduce:**

1. Visit https://build.origami.ft.com/v2/bundles/css?modules=o-header@^4,o-techdocs@^5,o-colors@^3,o-fonts@^2,o-forms@^2,o-buttons@^3 – this should output a working bundle

2. Copy the shrinkwrap URL (should be [something like this](/v2/bundles/css?modules=o-header%404.0.0%2Co-techdocs%405.0.1%2Co-colors%403.3.1%2Co-fonts%402.1.1%2Co-forms%402.0.0%2Co-buttons%403.1.1%2Co-autoinit%401.2.0&shrinkwrap=highlight.js%40undefined%2Clodash%403.10.1-npm%2Cftdomdelegate%402.0.3%2Co-assets%402.0.0%2Co-dom%402.0.0%2Co-ft-icons%403.3.3%2Co-grid%404.0.7%2Co-hierarchical-nav%403.1.0%2Co-hoverable%403.0.0%2Co-layers%402.0.1%2Co-squishy-list%402.0.0%2Co-viewport%402.0.0%2Csass-mq%403.2.9)) – this should output a 404 error, outlining that `danielditgens/highlight.js` doesn't exist

**Workaround (temporary)**

Replace `highlight.js%40undefined` in the shrinkwrapped URL with `Financial-Times/highlight.js-shim@5eb4727ba777e7683140de72ba01468e81639827` ([like this](https://build.origami.ft.com/v2/bundles/css?modules=o-header@4.0.0,o-techdocs@5.0.1,o-colors@3.3.1,o-fonts@2.1.1,o-forms@2.0.0,o-buttons@3.1.1,o-autoinit@1.2.0&shrinkwrap=ftdomdelegate@2.0.3,Financial-Times/highlight.js-shim@5eb4727ba777e7683140de72ba01468e81639827,lodash@3.10.1-npm,o-assets@2.0.0,o-dom@2.0.0,o-ft-icons@3.3.3,o-grid@4.0.7,o-hierarchical-nav@3.1.0,o-hoverable@3.0.0,o-layers@2.0.1,o-squishy-list@2.0.0,o-viewport@2.0.0,sass-mq@3.2.9)) – this should output a working bundle.

rowanmanning:
This is a far-from-ideal fix, because the shrink-wrapped URL for any
bundle that relies on Git sources isn't *technically* shrinkwrapped –
the HEAD of that source will be loaded.

So we're fixing the issue where an incorrect Git source is used for some
shrink-wrapped bundles, but we've uncovered a new issue: shrink-wrapped
bundles aren't always shrink-wrapped :/

I'll open a separate issue for that.

This fixes #26.

### DIFF
--- a/lib/moduleinstallation.js
+++ b/lib/moduleinstallation.js
@@ -292,6 +292,8 @@ ModuleInstallation.prototype = {
 					version: manifest.version,
 					target: manifest._target,
 					source: manifest._source,
+					originalSource: manifest._originalSource,
+					commit: (manifest._resolution ? manifest._resolution.commit : null)
 					// mainPathOverride can't be exposed here, because the install is shared between overriden and non-overriden sets
 				};
 			}
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -30,7 +30,13 @@ function _createFixedVersionsQueryString(components, filterMethod) {
 	return Object.keys(components).reduce(function(currentValue, componentName) {
 		const componentInfo = components[componentName];
 		const actualVersion = componentInfo.version;
-		const componentVersionTargetString = componentName + '@' + actualVersion;
+		let componentVersionTargetString;
+
+		if (componentInfo.originalSource.indexOf('/') !== -1) {
+			componentVersionTargetString = componentInfo.originalSource;
+		} else {
+			componentVersionTargetString = componentName + '@' + actualVersion;
+		}
 
 		// If the component matches a filter, don't include it in the list
 		if (filterMethod) {
--- a/test/integration/v2-bundles-css.js
+++ b/test/integration/v2-bundles-css.js
@@ -22,7 +22,7 @@ describe('GET /v2/bundles/css', function() {
 		});
 
 		it('should respond with the bundled CSS', function(done) {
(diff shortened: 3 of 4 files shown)


lib/moduleinstallation.js
```diff
@@ -292,6 +292,8 @@ ModuleInstallation.prototype = {
 					version: manifest.version,
 					target: manifest._target,
 					source: manifest._source,
+					originalSource: manifest._originalSource,
```
AlbertoElias: what's the difference with source?

rowanmanning: `_source` has been messed around with – I think it basically takes the last part of the URL or GitHub repo. So `financial-times/highlight.js` becomes `highlight.js`.

`_originalSource` is left as-is, and I use it for the new shrinkwrapping.

AlbertoElias:
Looks good!

PR #48, opened 2016-04-14
Update the Build Service hostname

rowanmanning:
:warning: **maybe don't merge until we've discussed as a team** :warning: 

This updates the hostname across the build service from `build.origami.ft.com` to `origami-build.ft.com`. I've made a few code changes as well to make this process simpler in future:

  * Moved all hostname references in the code to a single file
  * Reduce how much the tests rely on a specific hostname

There are some important gotchas to consider before we merge this:

#### Asset paths

Because of the way asset paths are added into the output bundles, every bundle will now request assets through Fastly rather than Akamai (even if the bundle itself is requested through Akamai).

E.g. a request to <code>http://<b>build.origami.ft.com</b>/v2/bundles/css?modules=o-icons</code> (Akamai-served) will contain the following font source (Fastly-served):

<pre>
@font-face{font-family:ft-icons;src:url(//<b>origami-build.ft.com</b>/files/o-ft-ico...
</pre>

This _shouldn't_ cause any issues, but it's a more broad test of Fastly than we were originally anticipating. We'll need a fair amount of testing in QA before we can confidently promote to production.

#### New URLs

Any new Build Service requests from this point onwards will go through Fastly, because all of the documentation has been updated to use the new URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ The Financial Times has published this software under the [MIT license][license]
 
 
 
-[build-service]: https://build.origami.ft.com/
+[build-service]: https://origami-build.ft.com/
 [ci]: https://circleci.com/gh/Financial-Times/origami-build-service
 [developer-spreadsheet]: https://docs.google.com/spreadsheets/d/1mbJQYJOgXAH2KfgKUM1Vgxq8FUIrahumb39wzsgStu0/edit#gid=0
 [docker-compose]: https://docs.docker.com/compose/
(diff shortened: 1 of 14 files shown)


AlbertoElias:
lgtm! Like the hostname module approach

rowanmanning:
Thanks Alberto :) going to wait and see what Platform Architecture says before merging this.

onishiweb:
👍

PR #49, opened 2016-04-18
Make the documentation hostname configurable

rowanmanning:
I don't know how I feel about this, it _could_ be a bit misleading to use Mustache-style syntax for this replacement. What do you think?

This is so we can quickly and easily switch between Fastly and Akamai in the documentation as well as the app. So I can safely deploy the change to the preferred hostname and roll back in seconds if anything goes wrong.


AlbertoElias:
I think Mustache template syntax is fine. lgtm

PR #50, opened 2016-04-26
Redirect deprecated routes

rowanmanning:
This redirects file, module, and documentation routes in the same way as bundle routes are redirected.

### DIFF
--- a/lib/express/v1-deprecated.js
+++ b/lib/express/v1-deprecated.js
@@ -1,0 +1,13 @@
+'use strict';
+
+const hostnames = require('../utils/hostnames');
+const redirectWithBody = require('../express/redirect-with-body');
+const URL = require('url');
+
+module.exports = function(req, res) {
+	const redirectUrl = '/v2' + URL.parse(req.url).path.replace(/^\/v1/, '');
+	const redirectBody = 'This endpoint has been deprecated. You are being redirected to ' + redirectUrl + '\nSee https://' + hostnames.preferred + '/v2/#api-reference for more information.\n';
+	res.type('txt');
+	res.header('Access-Control-Allow-Origin', '*');
+	redirectWithBody(res, 301, redirectUrl, redirectBody);
+};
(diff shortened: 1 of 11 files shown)


lib/express/v1-deprecated.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
AlbertoElias: what about putting this in the /deprecated folder?

lib/express/v1-deprecated.js
```diff
@@ -0,0 +1,13 @@
...
+
+const hostnames = require('../utils/hostnames');
+const redirectWithBody = require('../express/redirect-with-body');
+const URL = require('url');
+
+module.exports = function(req, res) {
+	const redirectUrl = '/v2' + URL.parse(req.url).path.replace(/^\/v1/, '');
+	const redirectBody = 'This endpoint has been deprecated. You are being redirected to ' + redirectUrl + '\nSee https://' + hostnames.preferred + '/v2/#api-reference for more information.\n';
```
AlbertoElias: more than deprecatated, this is now removing it

lib/express/v1-deprecated.js
rowanmanning: Good idea, I've moved it

lib/express/v1-deprecated.js
rowanmanning: Good point. I've tweaked the messaging

test/integration/v1-files.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const request = require('supertest');
+
+describe('GET /v1/files', function() {
```
AlbertoElias: I'd personally find it easier if this was in the files.js but in a different describe, but I don't have a strong opinion about it

AlbertoElias:
Looks good! I don't think this should break in any way, files and modules haven't changed

test/integration/v1-files.js
rowanmanning: Ah I'm trying to match the filenames in here to the URLs

AlbertoElias: Yup, and that's fine as well, just wanted to mention my personal preference :P

PR #53, opened 2016-05-03
Fix an error that occurs if modules are badly malformed

rowanmanning:
This addresses: https://app.getsentry.com/nextftcom/build-service-prod/issues/125987413/


AlbertoElias:
👍

PR #55, opened 2016-05-11
Remove the redirects for v1 files/modules

rowanmanning:
This caused issues with requests that are proxied and have different
paths. We've fixed this in a few places, but it highlights that we don't
have the full picture of how people are using/accessing the build
service.

This commit can be reverted once we've done some more comms around
moving to v2 routes, and have seen a significant drop in requests to v1
and unversioned routes.


onishiweb:
LGTM 👍

PR #56, opened 2016-05-13
Deprecation Documentation

rowanmanning:
This adds documentation on both the API version deprecation, and hostnames/Akamai. It might be easier to read this by running the Build Service locally and viewing the rendered HTML. Actually you can probably just checkout this branch and open the HTML file in-browser.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -326,6 +326,124 @@ X-FT-Build-Info: 1b1ab000a9b5642f6b8726039f1e79477b57c103; Tue, 15 Nov 2012 08:1
 				</tr>
 				</table>
 
+
+				<h2 id="deprecation">Deprecation</h2>
+
+				<h3 id="deprecation-api-versions">API versions</h3>
+
+				<p>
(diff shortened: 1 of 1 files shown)


docs/index.html
```diff
@@ -326,6 +326,214 @@ <h3 id="get-v1-modules-module-version-">GET /v2/modules/<code>module</code>@<cod
...
 				</table>
 
+
+				<h2 id="deprecation">Deprecation</h2>
+
+				<h3 id="deprecation-api-versions">API versions</h3>
+
+				<p>
+					All Build Service endpoints are versioned, and the current version is <code>v2</code>.
+					When deprecating old versions of the API we give at least three months notice,
```
alicebartlett: It would be good to link to the relevant bit of the Origami spec here i think

rowanmanning: Good idea!

alicebartlett:
This is  great, we should have done this ages ago!

PR #59, opened 2016-06-21
Babel disable

### LINKED ISSUE #58: Surface the babel-runtime disabling flag from OBT
### Problem
I want to use polyfill from the Polyfill Service and not the ones given by OBS. Now that the Polyfill Service is adding more ES6 polyfills we are seeing clashes between the polyfills from OBS and the ones from the Polyfill Service. Currently we can sidestep this issue by excluding these polyfills from the Polyfill Service and using the polyfills from OBS however this means that our end-users lose the benefit of the polyfill service as every single user will receive the polyfills given by OBS instead of only the users which require them.

### Solution
Have a query parameter which when set, disables the babel-runtime set of polyfills. OBT has had a flag to [disable the babel-runtime](https://github.com/Financial-Times/origami-build-tools/pull/354/) for a few months, meaning this problem doesn't exist for developers who use OBT directly.

rowanmanning:
This adds a `babel` parameter to the `/v2/bundles/js` endpoint which can be used to switch off Babel polyfills. It maps to the [`babelRuntime`](https://github.com/Financial-Times/origami-build-tools/pull/354/) option in OBT, and can be set to `none` or another falsy value. It defaults to `true`.

I'm happy to change the name of the parameter if this one isn't descriptive enough, or there's another name that makes more sense.

This PR resolves #58.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -171,9 +171,9 @@ o-typography:
 						<td>Querystring</td>
 						<td><em>(Optional)</em> If present and set to 'none', suppresses minification.  Otherwise output will be minified automatically.</td>
 					</tr><tr>
-						<td><code>babel</code></td>
+						<td><code>polyfills</code></td>
 						<td>Querystring</td>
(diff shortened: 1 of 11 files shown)


AlbertoElias:
👍

JakeChampion:
LGTM 👍

PR #62, opened 2016-08-15
Correct health checks

rowanmanning:
The panic guides for memory/disk usage were out of date.
Also, now that we're no longer using Alpine Node on Docker,
we need to revert back to the regular diskspace module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "bunyan": "^1.5.1",
     "cheerio": "^0.17.0",
     "commander": "^2.3.0",
-    "diskspace": "git+http://github.com/Financial-Times/diskspace.js.git#alpine",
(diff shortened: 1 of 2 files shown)


onishiweb:
LGTM

package.json
```diff
@@ -25,7 +25,7 @@
     "bunyan": "^1.5.1",
     "cheerio": "^0.17.0",
     "commander": "^2.3.0",
-    "diskspace": "git+http://github.com/Financial-Times/diskspace.js.git#alpine",
+    "diskspace": "git+http://github.com/Financial-Times/diskspace.js.git",
```
JakeChampion: I remember this issue when moving to Alpine :-)

Looks good to me!

PR #65, opened 2016-09-27
Modulemeta demo design guidance

onishiweb:
Add a new property to the module meta endpoint to return the design guidelines markdown file.

Required for the Registry's new demos construction.

### DIFF
--- a/lib/modulemetadata.js
+++ b/lib/modulemetadata.js
@@ -86,7 +86,6 @@ ModuleMetadata.prototype = {
 			}
 
 			// Saves Origami Registry hassle of finding and downloading the README.
-			// TODO:KL: instead of mixing text/markdown add type information or convert all to text/html
 			const readmeFilenames = ['README.md', 'readme.md', 'README', 'readme.txt', 'README.txt'];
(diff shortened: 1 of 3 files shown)


lib/modulemetadata.js
```diff
@@ -12,6 +12,21 @@ function ModuleMetadata(options) {
 }
 
 ModuleMetadata.prototype = {
+
+	_getFileFromList: Q.async(function*(filenames, module, installation) {
```
JakeChampion: Could we have tests for this new method?

onishiweb: Sure. However I have no idea where to start doing them. So some help would be appreciated.

JakeChampion: I believe this is the file you should add the tests to -- https://github.com/Financial-Times/origami-build-service/blob/284087007313a3decf2b45711d59a846141416b1/test/metadataapitest.js

onishiweb:
@JakeChampion @rowanmanning are we leaving the testing for now then?

JakeChampion:
Yep, the tests for this application need to have a look through as a whole.

PR #91, opened 2016-12-09
Log the original hostname used to access the app

rowanmanning:


lib/index.js
```diff
@@ -22,6 +23,17 @@ module.exports = function createApp(config) {
 	}
 
 	app.use(getBasePath);
+	app.use((request, response, next) => {
+		const originalUrl = request.headers['ft-original-url'];
```
JakeChampion: all vcls and akamais need to send this header

PR #98, opened 2017-01-23
Remove cli interface and use of uid and gid as heroku does not let you configure the user

JakeChampion:


JakeChampion:
Done @rowanmanning 👍

PR #104, opened 2017-01-27
Refactor registry module to make it simpler to understand

JakeChampion:
I hope the title is true


lib/registry.js
```diff
@@ -1,57 +1,44 @@
...
-			}
-		}.bind(this)).catch(function(err){
+const requestPromise = require('./utils/request-promise');
+
+class Registry {
+	constructor({
+		registryURL = 'http://registry.origami.ft.com'
+	} = {}) {
+		this.registryURL = registryURL;
+		// let packageList = this.refreshPackageList();
```
rowanmanning: Why commented and not removed?

PR #111, opened 2017-02-02
Moving closer to origami-service

JakeChampion:


test/integration/setup.js
```diff
@@ -1,32 +1,21 @@
...
 const uuid = require('uuid');
 
-before(function() {
+const tempdir = `/tmp/buildservice/tests-${uuid()}`;
 
-	// This overrides the Bower shorthand resolve, which allows us to
-	// refer to `mock-modules/<name>` in all of the tests. Bower will
-	// look on the local file-system instead of trying to load modules
-	// from GitHub or the registry.
-	process.env.bower_shorthand_resolver = `${__dirname}/{{shorthand}}`;
```
JakeChampion: ¯\_(ツ)_/¯

index.js
```diff
@@ -7,7 +7,7 @@ require('dotenv').load({
 });
 
 const process = require('process');
-const createApp = require('./lib/index');
+const buildService = require('./lib/index');
```
rowanmanning: This could be moved to `./lib/build-service` too. Or maybe we move all of them to `./lib/service`?

PR #112, opened 2017-02-13
Extracting middleware functions into their own modules

JakeChampion:


lib/routes/v1/index.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = app => {
+
+	// v1 endpoint redirect
+	app.get('/v1', (request, response) => {
+		const url = `${request.basePath}v2${request.url}`;
```
rowanmanning: Will this end up as `/base-path/v2/v1/rest-of-path`, as `v1` is not removed?

deprecated/v1-deprecated.js
```diff
@@ -6,12 +6,7 @@ const URL = require('url');
 
 module.exports = function(req, res) {
 	const requestPath = URL.parse(req.url).path;
-	let redirectUrl;
-	if (requestPath.indexOf('/v1') !== -1) {
-		redirectUrl = requestPath.replace(new RegExp('^/v1/'), req.basePath + 'v2/');
-	} else {
-		redirectUrl = requestPath.replace(new RegExp('^/'), req.basePath + 'v2/');
-	}
+	const redirectUrl = requestPath.replace(new RegExp('^/'), req.basePath + 'v2/');
```
JakeChampion: This will replace the first `/` in the path with `req.basePath + v2/`, no touching anything after the first `/`.

`req.basePath` has a default value of `/` which would make the start of the redirect url `/v2/`.

PR #121, opened 2017-04-04
Origami service error handling

rowanmanning:
This replaces the old error handling middleware with the Origami Service one. The repetition is reduced and we now send HTML errors.


lib/middleware/sanitize-errors.js
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		// User errors get a 400 status
+		if (error instanceof UserError || error.type === 'UserError') {
+			error.status = 400;
+			return next(error);
+		}
+
+		// Several different error codes indicate that a resource cannot be found
+		if (error.code === 'ENOTFOUND' || error.code === 'ENOENT' || error.code === 'ENORESTARGET') {
```
JakeChampion: Are these all 404s?

When do we get `ENORESTARGET`? If that is when a module exists but no version available in the semver range requested, should that be a 400 instead of a 404?

lib/middleware/sanitize-errors.js
```diff
@@ -0,0 +1,87 @@
...
+				return `by ${dependent.pkgMeta.name}@${dependent.pkgMeta._target}`;
+			});
+			return ` - Required at version ${pick.endpoint.target} ${requiredBy.join('\n')}`;
+		});
+		return `Cannot complete build: conflicting dependencies exist.\n\n${message}\n${picks.join('\n')}`;
+	}
+
+	// Compile errors need additional context
+	if (error.type === 'CompileError') {
+		return `Cannot complete build due to compilation error from build tools:\n\n${message}\n`;
```
JakeChampion: Do we have a stat for this in graphite?

rowanmanning: Yep, but it's done here: https://github.com/Financial-Times/origami-build-service/pull/121/files/20708030b12180a3f34edf52c077dda7b9268fe5#diff-cd08f18d1c9328a5440bdd89331c425aR18

lib/middleware/sanitize-errors.js
rowanmanning: I was going with what was already there, but I think I agree with it. A 404 indicates that the requested resource doesn't, and has never, existed. A 400 is a less specific user error and doesn't really convey as much IMO.

PR #126, opened 2017-04-28
Remove custom health checks and Origami Service

rowanmanning:
This uses Origami Service and the Health Check module, removing all of
the custom code.


lib/health-checks.js
```diff
@@ -1,127 +1,106 @@
...
+
+			// This check ensures that The Origami Registry
+			// is available on port 80. It will fail on a
+			// bad response or socket timeout
+			{
+				type: 'tcp-ip',
+				host: 'registry.origami.ft.com',
+				port: 80,
+				interval: 10 * 60 * 1000, // 10 minutes
+				id: 'github-tcp-port-80',
```
JakeChampion: copy paste mishap, should be registry not github

PR #134, opened 2017-07-03
Redirect all v1 and unversioned routes

rowanmanning:


deprecated/v1-deprecated.js
```diff
@@ -6,7 +6,7 @@ const URL = require('url');
 
 module.exports = function(req, res) {
 	const requestPath = URL.parse(req.url).path;
-	const redirectUrl = requestPath.replace(new RegExp('^/'), req.basePath + 'v2/');
+	const redirectUrl = requestPath.replace(new RegExp('^/(v1/)?'), req.basePath + 'v2/');
```
JakeChampion: What happens if the request is for `/v1`?

JakeChampion: Looks to be handled by another middleware

PR #142, opened 2017-09-28
Update origami-service

rowanmanning:


rowanmanning:
@JakeChampion: No shrinkwrap or lock for this, no. Was gonna update to Node.js 8 separately and add in a package-lock then.

PR #145, opened 2017-10-10
Allow only valid bower names to be used via our HTTP API

JakeChampion:
This stops relative paths from being used in ?modules=... as well as any other format which is not valid according to the bower.json name specification.

https://github.com/bower/spec/blob/master/json.md#name

If a module name in the modules query parameter is not valid, we return an HTTP 400 status code telling the user that the module name(s) is invalid.

### DIFF
--- a/lib/middleware/cleanModulesParameter.js
+++ b/lib/middleware/cleanModulesParameter.js
@@ -20,7 +20,7 @@ module.exports = function () {
 			*/
 			const moduleVersionPairs = request.query.modules.split(',').map(module => module.split('@'));
 
-			const invalidModuleVersionPairs = moduleVersionPairs.filter(function ([name, version = '']) {
+			const invalidModuleVersionPairs = moduleVersionPairs.filter(function ([name]) {
 				return !isValidBowerModuleName(name);
 			});
 
--- a/lib/middleware/cleanModulesParameter.js
+++ b/lib/middleware/cleanModulesParameter.js
@@ -1,0 +1,50 @@
+'use strict';
+
+const httpError = require('http-errors');
+
+module.exports = function () {
+
+	/**
+	 * Middleware used to enforce all modules names in the modules query parameter conform to the bower.json specification.
+	 * https://github.com/bower/spec/blob/master/json.md#name
+	 *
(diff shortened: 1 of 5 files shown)


lib/middleware/cleanModulesParameter.js
```diff
@@ -0,0 +1,50 @@
...
+				into [['o-colors', ''], ['o-grid', '^4'], ['o-techdocs', '*'], ['o-buttons', 'latest']]
+			*/
+			const moduleVersionPairs = request.query.modules.split(',').map(module => module.split('@'));
+
+			const invalidModuleVersionPairs = moduleVersionPairs.filter(function ([name]) {
+				return !isValidBowerModuleName(name);
+			});
+
+			if (invalidModuleVersionPairs.length > 0) {
+				const invalidModuleVersionsString = invalidModuleVersionPairs.map(([name]) => name).join(', ');
```
notlee: Couldn't this be just `const invalidModuleVersionsString = invalidModuleVersionPairs.join(', ');`?

lib/middleware/cleanModulesParameter.js
```diff
@@ -0,0 +1,50 @@
...
+	};
+};
+
+/**
+ * Checks a Bower module name conforms to the bower.json specification.
+ * @param {String} name A Bower module name.
+ * @returns {Boolean} Whether the name parameter is valid according to bower.
+ */
+function isValidBowerModuleName(name = '') {
+	return /^[a-z0-9_][a-z0-9_\.\-]*[a-z0-9_]$/.test(name.toLowerCase());
```
notlee: My regex skills are weak but I don't think these two cases are verified:
- Consecutive dashes or dots not allowed.
- 50 characters or less.
I don't however think that's an issue considering the problem you're solving.

notlee:
Looks good @JakeChampion ! 🙌

lib/middleware/cleanModulesParameter.js
JakeChampion: That is true, we should cater for those! Added an issue for that work https://github.com/Financial-Times/origami-build-service/issues/147

Issue #119, opened 2017-03-21
Clarification on polyfills


callumlocke:
Need clarification on this:

> `polyfills` (Optional) If present and set to 'none', does not add polyfills to the output. Use this if your bundle is conflicting with other polyfills (e.g. through the Polyfill Service).

Which polyfills are we talking about? What features do I need to get from polyfill service if I set `polyfills=none` on the build service?

JakeChampion:
Currently build service bundles the babel-polyfill set inside the JS bundle. If you set `polyfills=none`, you will not get the babel-polyfill set. The polyfills you need from polyfill service will vary upon your codebase. I.E. If you are using `Map`s, you will need the `Map` polyfill etc.

The main issue comes down to if you are using new syntax which relies on new JS features such as `for..of`, if using `for..of` you will need to polyfill the `@@iterator` methods on whatever Object types you are iterating over. `NodeList.prototype.@@iterator`, `Array.prototype.@@iterator` etc.

PR #150, opened 2017-11-07
Clarified the polyfill flag

Limess:
This did not go into any detail of what was loaded, added information and references for `babel-runtime` and the accompanying plugin.

Closes https://github.com/Financial-Times/origami-build-service/issues/119.


Limess:
Reworded this since approval as it made not-so-much sense @JakeChampion.

PR #164, opened 2018-02-13
Add a new endpoint for retrieving minimal demo HTML

rowanmanning:
⚠️ *This is a work-in-progress, there are no tests yet*

⚠️ *It may be easier to view this PR with whitespace ignored, as I cleaned up some mixed whitespace*

This endpoint outputs simple HTML as plain text, so that it can be
rendered as-is in browser. It replicates the behaviour of the current
Origami Registry in stripping out elements that are not required for the
demo. We'll be using this endpoint in the new registry.


lib/middleware/outputDemo.js
```diff
@@ -53,7 +54,49 @@ const validateOnlyOrigamiModules = (installation) => {
 		});
 };
 
-module.exports = app => {
+const extractMinimalHtml = (fullHtml) => {
+	const $ = cheerio.load(fullHtml);
+
+	// Grab the first demo chunk
+	const demoChunk = $('.o-registry-demo-chunk').first();
```
notlee: Did not know this was a thing! It's only used by `o-ft-header` and `o-tweet` _I think_ which are deprecated. It also appears that their demos don't work. I think we can remove this?

lib/middleware/outputDemo.js
```diff
@@ -53,7 +54,49 @@ const validateOnlyOrigamiModules = (installation) => {
...
+	}
+
+	// If no demo chunk was found, use the body
+	const body = $('body');
+
+	// Find all top-level non-template script and link elements
+	const elements = body.find('> script:not([type="text/template"]), > link');
+
+	// Strip script and link elements that reference Origami URLs
+	for (const el of elements.toArray()) {
```
JakeChampion: Why is toArray required? Is the set not iterable?

PR #165, opened 2018-04-17
Branding

notlee:
Adds brand support to the current version of the build service. To request branded components add a brand query parameter to a build service request of either "master", "internal", or "whitelabel". E.g. `/v2/bundles/css?modules=o-table@^6.7.0&brand=internal`

I need to figure out the best way to write a test this. 😬Would quite like a functional test to inspect `mainSassFileContents` for a brand variable, but could do an integration test than uses a new fake component. Thoughts?

kapture 2018-04-17 at 17 57 48


lib/middleware/cleanBrandParameter.js
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports = function () {
+	/**
+	 * Middleware used to enforce the brand query parameter is alphanumeric.
+	 */
+    return (request, response, next) => {
+        const brand = request.query.brand;
+        if (brand && !/^[a-z0-9]+$/i.test(brand)) {
+            next(
+                httpError(400, `The brand parameter must be alphanumeric only: ${brand}`)
```
wheresrhys: Should this, rather, 400 if brand isn't from a whitelist? It's unlikely brands are gonna proliferate... fo rthe time being at least

notlee: Yeah I think that's fair. If nothing else it'll help people identify typos in the brand param.

PR #188, opened 2019-01-22
Report to Raven when a component does not support the requested brand.

notlee:
**goal**

SCSS compilation fails for `o-layout@v3.0.0-beta-1` and the master brand. We don't intent to fix it, as o-layout only supports the internal brand. 

Instead of a random SCSS compilation error:
>cannot complete build due to compilation error from build tools: .../bower_components/o-typography/src/scss/_functions.scss Error: Invalid null operation: "null times 0.87". on line 21 of  [...]

OBS should throw a more helpful error:
>o-layout does not support the master brand. Please set a supported brand (internal) using the "brand" query parameter, otherwise contact the Origami team to propose adding master brand support to o-layout.

**gotchas**

1. o-layout < v3 mistakenly set the brand SCSS variable itself, therefore [some users](https://financialtimes.splunkcloud.com/en-US/app/search/search?s=&q=search%20index%3D%22heroku%22%20%22host%3Dorigami-build-service%22%20method%3DGET%20(path%3D%22*css*%22%20AND%20path%3D%22*o-layout*%22%20AND%20path!%3D%22*internal*%22%20AND%20path!%3D%22*demo*%22)&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-1d&latest=now) have not have set their brand parameter in the OBS request. No error should be thrown in this case.
2. Brand support has been added to components incrementally, therefore any components which are not branded have to assume support for any brand.
3. Some components like `o-colors` have added support for different brands at different times. It's possible that a branded component has been used for an unsupported brand. 

**approach**

Throwing an error when a component does not support the requested brand is technically a breaking due to gotcha 3. I think it's a reasonable error to throw though if we can confirm it won't break our users builds. This PR will report to Sentry any builds which would fail. If none are reported, we may assume we can safely throw an error for real (I suggest at least a month, to allow the max-age of builds to expire; a month will also allow for sites with little traffic to make a request, as branding is quite recent this concerns me less). 👈 Feedback on whether this is a dreadful idea appreciated.

### DIFF
--- a/lib/cssbundler.js
+++ b/lib/cssbundler.js
@@ -175,7 +175,7 @@ CssBundler.prototype = {
 		return readpromises;
 	},
 
-	_compileSassFiles: Q.async(function* (installation, requestedComponents, allComponents) {
+	_compileSassFiles: Q.async(function* (installation, requestedComponents, allComponents, brand) {
 		const variables = {};
 
 		if (brand) {
--- a/lib/cssbundler.js
+++ b/lib/cssbundler.js
@@ -211,7 +211,7 @@ CssBundler.prototype = {
 				let brands = [];
 				if (isRequested) {
 					const manifest = yield installation.getOrigamiManifest(name);
-					supportsBrand = manifest && !manifest.brands || manifest.brands.includes(brand);
+					supportsBrand = manifest && (!manifest.brands || manifest.brands.includes(brand));
 					brands = manifest && manifest.brands ? manifest.brands : [];
 				}
 
--- a/lib/cssbundler.js
+++ b/lib/cssbundler.js
@@ -9,6 +9,7 @@ const CompileError = require('./utils/compileerror');
 const streamCat = require('streamcat');
 const hostnames = require('./utils/hostnames');
 const HTTPError = require('./express/httperror');
+const raven = require('raven');
 
 const gulp = require('gulp');
 const obt = require('origami-build-tools');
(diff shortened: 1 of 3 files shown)


lib/cssbundler.js
```diff
@@ -195,6 +196,38 @@ CssBundler.prototype = {
 				const comp = requestedComponents[name] || allComponents[name];
 				const isRequested = name in requestedComponents;
 
+				// Check o-layout < 3.0.0 is not installed.
+				// If it is allow the build regardless of the set brand.
+				// This is because o-layout < 3.0.0 mistakenly set the brand in its SCSS `$o-brand: internal`.
+				// Therefore projects may have included it without setting the brand.
+				let oLayoutException = false;
+				if (allComponents.hasOwnProperty('o-layout')) {
+					oLayoutException = parseFloat(allComponents['o-layout'].version) < 3;
```
JakeChampion: Would parseFloat work with a SemVer version such as 2.2.2?

JakeChampion:
Is it possible to have some tests to cover this feature?

lib/cssbundler.js
notlee: It would to confirm it's less than 3.

notlee:
>Is it possible to have some tests to cover this feature?
@JakeChampion My intention is to add an integration test if we keep it (I'll create an issue with details).

PR #200, opened 2019-08-27
 Add source query parameter.

notlee:
**Overview**
 Add source query parameter:

- Optional until v3, falls back to "build-service" when not set.
- Used to set a global `$system-code` variable in Sass builds.
- Refactors a little to pass `brand` and `source` variables in the
builds `options` object, rather than a separate argument.

The `$system-code` Sass variable will be used by Origami components 
to set a source parameter on image service requests.

**Relates to:**
https://github.com/Financial-Times/origami-build-tools/pull/604
https://github.com/Financial-Times/o-footer-services/issues/10 
(and similar component issues)

### DIFF
--- a/lib/democompiler.js
+++ b/lib/democompiler.js
@@ -54,7 +54,7 @@ DemoCompiler.prototype = {
 				dist: true,
 				demoFilter: options.demoName,
 				cwd: installation.getComponentDir(options.moduleName),
-				brand: options. brand
+				brand: options.brand
 			};
 
 			log.trace(demoConfig, 'Starting OBT Demo build');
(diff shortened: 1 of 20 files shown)


lib/democompiler.js
```diff
@@ -47,14 +47,14 @@ DemoCompiler.prototype = {
...
-	_doBuild: function (installation, moduleset, brand, options) {
+	_doBuild: function (installation, moduleset, options) {
 		return new Promise((resolve, reject) => {
 			const log = this.log;
 			const demoConfig = {
 				dist: true,
 				demoFilter: options.demoName,
 				cwd: installation.getComponentDir(options.moduleName),
-				brand: brand
+				brand: options. brand
```
notlee:
```suggestion
				brand: options.brand
```

PR #216, opened 2019-11-22
Pin unversioned modules to the current major

rowanmanning:
We have encountered an issue with the current major cascade, in that the
number of Build Service URLs which include unversioned modules is much
higher than we thought.

This means that as we release new major versions, Build Service bundles
are breaking all over the place. After some discussion, we decided that
the Build Service v2 should pin unversioned modules to the previous
major version and Build Service v3 should _require_ a version to be
specified.

This PR achieves this in the quickest way possible, but this should
probably be revisited after the cascade is over.


lib/data/default-module-versions.json
```diff
@@ -0,0 +1,3 @@
+{
+	"o-buttons": "^3.0.0"
```
chee: why 3? i realise this is a hotfix but it might be nice to have comments explaining why the default versions are chosen

rowanmanning: Sorry this is not ready yet, we're waiting to get all the real versions and this was just a temporary test number

PR #221, opened 2020-01-28
Update cache purge documentation

notlee:
We hope to automate this in a future major version of the build
service. https://github.com/Financial-Times/origami-build-service-v3/issues/64

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,40 @@ You can promote either through the Heroku interface, or by running the following
 make promote
 ```
 
+You may need to [clear cdn cache](#cache-purge) for the release to take immediate effect. For example when updating documentation.
+
+Cache Purge
+----------
+
+First, change the hostname in your request to `origami-build-service-eu.herokuapp.com`. If your update does not appear, an old version is cached on the file system. Clear this by restarting the Heroku dynos:
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,6 +114,40 @@ You can promote either through the Heroku interface, or by running the following
...
+Cache Purge
+----------
+
+First, change the hostname in your request to `origami-build-service-eu.herokuapp.com`. If your update does not appear, an old version is cached on the file system. Clear this by restarting the Heroku dynos:
+
+```sh
+heroku restart --app origami-build-service-eu
+```
+
+If your change does appear then the old result may be cached by our CDN. You'll need to wait for a while, or clear the CDN cache. To clear CDN cache login to Fastly and find the `ft.com` service. Clear a specific URL (e.g. for a documentation update) or one or more of the following [surrogate keys](https://docs.fastly.com/en/guides/getting-started-with-surrogate-keys):
```
JakeChampion: The origami build service fastly service is the one to purge from

notlee: Thanks! I always get that the wrong way around 😳 All the more reason to document! 

Inner monologue: I always get this wrong. As origami build service fastly service is the obvious right answer, it must be the ft.com fastly service!

PR #223, opened 2020-01-31
Bump origami-build-tools version and other dependencies.

notlee:
origami-build-tools@5.8.2 does not add the `default` polyfill
to demo pages.

Otherwise this can give the false impression a component works
in a browser without polyfills. Only polyfills requested in
`origami.json` (`browserFeatures.required`) should be included.

https://github.com/Financial-Times/origami-build-tools/issues/622


notlee:
I'll save this for a Monday release

PR #306, opened 2020-06-30
Remove n-alert-banner

andygout:
[`n-alert-banner`](https://github.com/Financial-Times/n-alert-banner) will shortly be archived.

Once this has happened this PR will be ready for review and merge.


notlee:
We need to keep this so Build Service requests continue to work as expected. Hopefully if you've found all the other `n-alert-banner` there will be no build service requests :)

PR #383, opened 2020-10-16
Add middleware to ensure all requested Origami modules are built using the same Origami Specification version

JakeChampion:
This uses a fixed list (json file) of Origami Components to check if something is an Origami Component.

If the requested module is an Origami Component and is using a Git reference instead of a SemVer version, we assume they are requesting a version of the Origami Component which was built using Origami Spec v1.


lib/middleware/checkModulesAllUseSameOrigamiVersion.js
```diff
@@ -0,0 +1,90 @@
...
+function getOrigamiVersion(name, version) {
+	if (origamiVersion1Modules[name]) {
+		// If the version is not a range, it is likely to be a git ref such as a branch name.
+		// If the component has a version which was origami spec v1, then we assume the
+		// git ref is going to return a version of the component which is origmai spec v1.
+		if (!semver.validRange(version)) {
+			return 1;
+		}
+
+		if (semver.intersects(origamiVersion1Modules[name], version)) {
```
JakeChampion, line 24: What we want is to confirm that the requested version range will _always_ return a version of the component which was built against spec v1.

`semver.intersects` does not do that, it will return `true` if the two ranges have some intersection, which means we could give it a range which can resolve to a version of the component which was built against spec v2 and it would return `true`

```
// intersects(range): Return true if any of the ranges comparators intersect
semver.intersects('<= 4', '>= 4') // returns true
```

I believe what we want is to confirm that the requested version range is a subset of the range which contains all origami spec v1 versions of the component.

```
//subset(subRange, superRange): Return true if the subRange range is entirely contained by the superRange range.
semver.subset('>=4', '<=4') // returns false
```

```suggestion
		if (semver.subset(version, origamiVersion1Modules[name])) {
```

lib/middleware/checkModulesAllUseSameOrigamiVersion.js
```diff
@@ -0,0 +1,91 @@
...
+function getOrigamiVersion(name, version) {
+	if (origamiVersion1Modules[name]) {
+		// If the version is not a range, it is likely to be a git ref such as a branch name.
+		// If the component has a version which was origami spec v1, then we assume the
+		// git ref is going to return a version of the component which is origmai spec v1.
+		if (!semver.validRange(version)) {
+			return 1;
+		}
+	//TODO: What should we do if the requested version intersects with versions which are spec v1 but is not a full subset?
+	// E.G. If component `o-no` versions upto and including v4 are built against origami spec v1 and from v5 onwards is built againt origami spec v2 it's version range which covers spec v1 is `<= 4`. If the version range being requested is `>= 4`, what should we do? This range can return a spec v1 or a spec v2 version and the result is dependent upon whether any other modules in the dependency tree require of `o-no`.
```
notlee, line 24: Do we need to enforce all requested modules use the same Origami version? We could install a mix and write a main file accordingly? I suppose if we do that will make Build Service maintaince easier and make it easier to deprecate v1.

If we are enforcing only components of the same spec can be used together via the build service, we should probably error if there's a range which spans both?

JakeChampion: I think we need to enforce a difference between spec v1 and spec v2 components because spec v2 components have different requirements such as they only support dart-sass.

JakeChampion: I like the idea to error if the range could span both 👍

JakeChampion: Done this in https://github.com/Financial-Times/origami-build-service/pull/383/commits/577eb1e08692ac90db0fed240596d549b7bde89c

test/unit/lib/middleware/checkModulesAllUseSameOrigamiVersion.test.js
```diff
@@ -0,0 +1,240 @@
...
+				it('calls `next`', () => {
+					assert.calledOnce(spyNext);
+					assert.calledWithExactly(spyNext);
+				});
+			});
+
+			describe('when the \'modules\' query string contains only modules with versions which use origami spec v2', () => {
+
+				beforeEach(() => {
+					origamiService.mockRequest.query.modules = 'o-teaser@^4.0.0';
```
notlee, line 60: `o-teaser@^4.0.0` is Origami spec v1 not v2? The latest is v5, I guess for testing purposes it would be safe to go with v6 which does not exist yet

JakeChampion:
```suggestion
					origamiService.mockRequest.query.modules = 'o-teaser@^6.0.0';
```

notlee (approved):

JakeChampion:
I'm going to close this as we changed our minds and decided to make a v3 of the build service which is for origami spec v2 components and the already existing v2 build service will only support v1 origami components which is already the case

PR #394, opened 2020-11-05
Deploy pull-requests to the dev app and not the staging app

JakeChampion:
🙈 luckily we do not use the heroku pipeline promotion feature when deploying to production because it we did we would have been deploying random pull-requests to production!


chee (approved):
lol!

PR #404, opened 2020-12-09
Do not crash the app if the NPM cache fails to prime.

notlee:
This might happen if a component is not published for some reason.
The Build Service failed recently because of this:
https://financialtimes.slack.com/archives/C02FU5ARJ/p1607506789255200
`@financial-times/o-fonts-assets` was published to the test npm
registry but not to the official npm registry.

When v2 components are on the official npm registry we can be
confident that our tests will catch a future regression like this.
Supports: Financial-Times/origami#86

The PR also includes some dependency bumps including node-sass v5 (see separate commits):
- Only support LTS and current Node versions
- Remove deprecated process.sass API


notlee:
@JakeChampion tested this on the dev app without the `NPM_REGISTRY_URL` environment variable. Looks good!

index.js
```diff
@@ -40,9 +40,9 @@ process.env.HOME = options.tempdir; // Workaround: Bower ends up using $HOME/.lo
 buildService(options)
 	.listen()
 	.then(() => {
-		return primeNpmCache(options.npmRegistryURL);
+		return primeNpmCache(options.npmRegistryURL).catch(console.log);
```
JakeChampion, line 43: Let's use console.error so that it goes to stderr instead 👍 
```suggestion
		return primeNpmCache(options.npmRegistryURL).catch(console.error);
```

JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v355.1.1 :tada:

PR #408, opened 2020-12-14
Add v3/files endpoint 

JakeChampion:
This adds the files endpoint to the version 3 of the API, which supports npm instead of bower (origami v2 components).

Differences from version 2 files endpoint:

- version 2 only used uri path parameters for the api and not any query parameters (`GET /v2/files/module@version/path` whereas version 3 uses only query parameters and not path parameters (`GET /v3/files?module=...&file=...&system_code=...`)

- version 3 requires a system_code query parameter to be present, which is the biz-ops system code for whatever is making the API call

- version 3 uses the npm registry instead of the bower registry

Screenshot of the documentation -- https://origami-build-service-dev.herokuapp.com/v3/docs/api#get-v3-files
image

Screenshot of the migration notes --
image

### DIFF
--- a/lib/middleware/v3/createJavaScriptBundle.js
+++ b/lib/middleware/v3/createJavaScriptBundle.js
@@ -13,6 +13,7 @@ const { parseCallbackParameter } = require('./parseCallbackParameter');
 const { createPackageJsonFile } = require('./createPackageJsonFile');
 const { installDependencies } = require('./installDependencies');
 const { bundleJavascript } = require('./bundleJavascript');
+const { parseSystemCodeParameter } = require('./parseSystemCodeParameter');
 
 const rimrafOptions = {
 	glob: false,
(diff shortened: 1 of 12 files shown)


notlee:
Me and Jake discussed this:
- The main usecase for this endpoint is to include font files from `o-fonts-assets`. 
- Other uses appear to be old component assets ([file endpoint Splunk statistics](https://financialtimes.splunkcloud.com/en-GB/app/search/search?sid=1608031697.3000470)). 
- No current component version includes assets. None will in the same way going forward, as it will not work with npm. There's [a proposal for handling component assets](https://github.com/Financial-Times/origami/pull/102) in the future.
- If we require a system code to include fonts, as pass through the system code of the project consuming Origami, the client won't cache fonts between pages of the ft (pretty major, fonts are heavy)
- Instead we decided it would be better to remove the file endpoint from v3 and host fonts in S3. Perhaps we go ahead and [implement the asset proposal for all components](https://github.com/Financial-Times/origami/pull/102)?

Issue #399, opened 2020-11-30
Integrate with Biz-Ops via the NodeJS API Client


chee:
- [ ] Integrate with Biz-Ops via the NodeJS API Client -- https://github.com/Financial-Times/biz-ops-client

_Originally posted by @chee in https://github.com/Financial-Times/origami-build-service/pull/391#r532678099_

PR #417, opened 2021-01-07
Validate system-codes with the new http api

JakeChampion:
This changes how we validate the system_code query parameter, previously we were checking it adhered to the same regular expression that is currently used within the Biz-Ops project. This approach was not complete because it never proved that the provided system_code value was in fact a real system code stored within Biz-Ops, which is ideally what we want to do.

There is a new HTTP API available to validate whether something is a real system code within Biz-Ops, https://system-codes.in.ft.com/ - This is what we are now using instead of the regular expression.

The HTTP API is implemented entirely within Fastly VCL, which means it is fast to respond and very unlikely to go offline. If it does take a long time to respond or does infact go offline we will 'fail open' and pretend that the provided system_code value is a real valid system code. This ensures that we don't break customers websites simpliy because we can't validate their provided system_code.

This resolves https://github.com/Financial-Times/origami-build-service/issues/399

### DIFF
--- a/lib/middleware/v3/parseSystemCodeParameter.js
+++ b/lib/middleware/v3/parseSystemCodeParameter.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const UserError = require('../../utils/usererror');
+const axios = require('axios').default;
 
 module.exports = {
     parseSystemCodeParameter: parseSystemCodeParameter
@@ -14,9 +15,9 @@ module.exports = {
  *
  * @param {string} systemCode The system-code to compile the css for.
  * @throws {import('../../utils/usererror')}
- * @returns {string} The system-code
+ * @returns {Promise<string>} The system-code
  */
-function parseSystemCodeParameter(systemCode) {
+async function parseSystemCodeParameter(systemCode) {
 	if (typeof systemCode !== 'string') {
 		throw new UserError('The system_code query parameter must be a string.');
 	}
@@ -25,7 +26,7 @@ function parseSystemCodeParameter(systemCode) {
 		throw new UserError('The system_code query parameter can not be empty.');
 	}
 
-	if (isValidSystemCode(systemCode)) {
+	if (await isValidSystemCode(systemCode)) {
 		return systemCode;
 	} else {
 		throw new UserError('The system_code query parameter must be a valid Biz-Ops System Code.');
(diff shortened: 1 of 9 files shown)


notlee:
👍 nice

> it does take a long time to respond or does infact go offline we will 'fail open' and pretend that the provided system_code value is a real valid system code. This ensures that we don't break customers websites simpliy because we can't validate their provided system_code.

👍 nice 

🤚 We should treat this as a breaking change because it could break websites, even if there's just a typo in otherwise correct useage. Rebase to v3 and release with that?

JakeChampion:
> 🤚 We should treat this as a breaking change because it could break websites, even if there's just a typo in otherwise correct useage. Rebase to v3 and release with that?

This is already only applied to the v3 api

notlee:
Sorry @JakeChampion I missed that!

notlee (approved):

JakeChampion:
> Sorry @JakeChampion I missed that!

I should have mentioned v3 in the commit and pr description, it's my bad

lib/middleware/v3/parseSystemCodeParameter.js
```diff
@@ -36,15 +37,33 @@ function parseSystemCodeParameter(systemCode) {
...
+			url: 'https://system-codes.in.ft.com/v1/check',
+			method: 'get', // default
+			params: {
+				systemCode: code
+			},
+			timeout: 1000 * 10,
+			responseType: 'json'
+		});
+
+		if (response.status === 200) {
```
notlee, line 54: Pedantic non-blocking suggestion: this could be rewritten to avoid nesting?

JakeChampion: Could you add a suggestion codeblock with how you would write it, I can't figure out how you would avoid the nesting here

lib/middleware/v3/parseSystemCodeParameter.js
```diff
@@ -36,15 +37,33 @@ function parseSystemCodeParameter(systemCode) {
...
+			if (response.data.exists === true) {
+				return true;
+			} else {
+				return false;
+			}
+		} else {
+			return true;
+		}
```
notlee, line 62:
```suggestion
		if (response.status === 200 && response.data.exists !== true) {
			return false;
		}

		return true;
```

lib/middleware/v3/parseSystemCodeParameter.js
notlee: Of course! I dont know why I didn't straight away. I had to do it in [a new multiline comment](https://github.com/Financial-Times/origami-build-service/pull/417/files#r553479019)

JakeChampion: Thanks, this was done in [`28c8476` (#417)](https://github.com/Financial-Times/origami-build-service/pull/417/commits/28c84768ced55c0849c9ab33e7d6d775cdf3e2b5)

PR #426, opened 2021-01-20
Add /v3/font api endpoint

JakeChampion:
This adds an API endpoint which can be used to serve font files. The reason for adding this new API endpoint is because the old /v2/files API endpoint which was used to serve any file from a repository was only used for serving the font files from o-fonts-assets.

This currently will not work until we start publishing o-fonts-assets to an npm registry. We can't publish o-fonts-assets to as a public package on an npm registry because that would violate our purchase agreement on the fonts. We need to publish the package to either a private registry or as a private package on a public registry.

Here is an example API call:
`/v3/font?version=^1&font_name=BentonSans-Light&font_format=woff2&system_code=origami-registry-ui`


JakeChampion:
The tests pass when using an npm registry which has o-fonts-assets on it such as https://origami-npm-registry-prototype.herokuapp.com/

notlee:
>An alternative API I thought of was to not take a filepath and instead take a font-name and a font-file-format like this: [...]

I like this format for the reasons you outlined 👍 It also means the api isn't tied to the folder structure of o-fonts-assets

JakeChampion:
> > An alternative API I thought of was to not take a filepath and instead take a font-name and a font-file-format like this: [...]
> 
> I like this format for the reasons you outlined 👍 It also means the api isn't tied to the folder structure of o-fonts-assets

My implementation would have tied the folder structure to o-fonts-assets 😅.

Do you have an implementation idea which doesn't tie it to the folder structure?

notlee:
>Do you have an implementation idea which doesn't tie it to the folder structure?

No, I meant the public api :)

index.js
```diff
@@ -22,7 +22,7 @@ const options = {
 		process.env.BOWER_REGISTRY_URL || 'http://origami-bower-registry.ft.com',
 	tempdir: `/tmp/buildservice-${process.pid}/`,
 	testHealthcheckFailure: process.env.TEST_HEALTHCHECK_FAILURE || false,
-	npmRegistryURL: process.env.NPM_REGISTRY_URL || 'https://registry.npmjs.com'
+	npmRegistryURL: process.env.NPM_REGISTRY_URL || 'https://origami-npm-registry-prototype.herokuapp.com/'
```
notlee, line 25: we can put this back now?
```suggestion
	npmRegistryURL: process.env.NPM_REGISTRY_URL || 'https://registry.npmjs.com'
```

lib/middleware/v3/outputFont.js
```diff
@@ -0,0 +1,111 @@
...
+	readdir: gracefulFs.readdir,
+	readdirSync: gracefulFs.readdirSync
+};
+
+/**
+ * @param {import('express').Request} request
+ * @param {import('express').Response} response
+ */
+const outputFont = async (request, response) => {
+	await fs.mkdir('/tmp/bundle/', {recursive: true});
```
notlee, line 48: We don't use `/tmp/bundle/` because a true temporary directory  `bundleLocation` created after?
```suggestion
```

lib/middleware/v3/parseFontFormatParameter.js
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @param {string} [fontFormat] The format of the font to be returned.
+ * @throws {import('../../utils/usererror')}
+ * @returns {string} The format of the font to be returned.
+ */
+const parseFontFormatParameter = fontFormat => {
+	if (fontFormat === undefined) {
+		throw new UserError('The font_format query parameter must be a string.');
+	}
+
```
notlee, line 15:
```suggestion
```

This is handled by the second conditional

lib/middleware/v3/parseVersionParameter.js
```diff
@@ -0,0 +1,39 @@
...
+	if (typeof version !== 'string') {
+		throw new UserError('The version query parameter must be a string.');
+	}
+
+	if (version !== version.trim()) {
+		throw new UserError(`The version query parameter cannot have whitespace at the start or end. The provided version was "${version}"`);
+	}
+
+	if (! semver.validRange(version)) {
+		throw new UserError(`The version ${version} is not a valid version.\nPlease refer to TODO (build service documentation) for what is a valid version.`);
```
notlee, line 30: Will obs be protected from reflected XSS when this error makes it to the error page template?

views/migration.html
```diff
@@ -15,11 +15,16 @@ <h2 id="v2-to-v3-migration">Migrating from v2 to v3</h2>
...
+
+    <h3>CSS endpoint</h3>
+
     <p>Version 3 no longer uses LibSass for compiling Sass and instead uses <a href="https://sass-lang.com/dart-sass">DartSass</a>.</p>
 
-    <p>Version 3 no longer uses Webpack for bundling JavaScript and instead uses <a href="https://esbuild.github.io/">ESBuild</a>.</p>
+    <h3>JS endpoint</h3>
 
-    <p>Version 3 has removed the <code>/modules</code> endpoint.</p>
+    <p>Version 3 no longer uses Webpack for bundling JavaScript and instead uses <a href="https://esbuild.github.io/">ESBuild</a>.</p>
```
notlee, line 27: Not for this PR but I guess my question from a user point of view is "so what?" We can probably delete this from the migration guide if we don't expect users to need to take an action

notlee (commented):
Nice work!

lib/middleware/v3/outputFont.js
JakeChampion: We need this line otherwise the next line will throw an error saying `/tmp/bundle` does not exist.

This line creates the `/tmp/bundle` directory, the next line creates a folder within `/tmp/bundle` with a random name 👍

notlee: I misread that, thanks :)

lib/middleware/v3/parseVersionParameter.js
JakeChampion: Good question, it was susceptible as the error page set it's content-type to html. I've changed it to now set the content-type to plaintext, which is not vulnerable to XSS -- daf4662426867a5878449c69c91eabe522b780e4

notlee (approved):

notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v357.1.0 :tada:

PR #437, opened 2021-02-02
Install `o-fonts-assets` with Github for the v3 files endpoint

notlee:
We install from the Github repository because `o-fonts-assets` is
not available via npm as it must remain private for license reasons.

With our current Github license we could publish `o-fonts-assets`
to the Github packages registry privately. However we would have
to use a different scope to `@financial-times` so that we can install
from both the Github and npm registry. Unless we were to move all
`@financial-times` packages to the Github packages registry. We
think the later would be the best approach but would take time
to propose and implement across the ft. Hence for now we opted
to install the Github repository directly.

Update from Jake:
>There is also the third option of waiting to see if github/npm will 
>solve this for us via https://github.com/npm/rfcs/issues/275 or 
>https://github.com/npm/feedback/discussions/87

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,9 +17,12 @@ jobs:
     - run: make test-unit-coverage
       env:
         ORIGAMI_GITHUB_TOKEN: ${{ secrets.ORIGAMI_GITHUB_TOKEN }}
+        DEBUG_SAFE_TO_LOG: ${{ secrets.DEBUG_SAFE_TO_LOG }}
     - run: make test-integration
       env:
         ORIGAMI_GITHUB_TOKEN: ${{ secrets.ORIGAMI_GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


JakeChampion:
> Unless we were to move all
> `@financial-times` packages to the Github packages registry.

There is also the third option of waiting to see if github/npm will solve this for us via https://github.com/npm/rfcs/issues/275 or https://github.com/npm/feedback/discussions/87

.github/workflows/test.yml
```diff
@@ -12,5 +12,11 @@ jobs:
     - run: npm ci
     - run: make verify
     - run: make test-unit-coverage
+      env:
+        GITHUB_TOKEN: ${{ secrets.ORIGAMI_GITHUB_TOKEN }}
```
JakeChampion, line 16: This won't work as that namespace is restricted to github -- https://docs.github.com/en/actions/reference/environment-variables#naming-conventions-for-environment-variables

JakeChampion (approved):

PR #445, opened 2021-02-12
Add URL Updater

notlee:
The URL Builder accepts an existing Build Service URL and builds
an updated URL which includes the latest version number for all
components, helping users to upgrade their Build Service requests
and migrate to the new version of components. To further help a
table is displayed which shows the components which were requested,
the latest version number for the component, a label to show if it
is up to date or not, and a link to the readme to access the
migration guides.

So far it does not handle all cases, e.g:
- Only the "bundles" endpoint is supported
- Module filepaths and non-Origami components aren't supported
- If the latest version of components can only be used with the
(upcoming) v3 Build Service endpoint then the generated URL will
include a v3 Build Service url but will not warn about any
parameters used which are  no longer be supported in v3 of the
Build Service

### Screenshots

1 out of date module:
Screenshot_2021-02-12 Origami Build Service

modules up to date:
Screenshot_2021-02-12 Origami Build Service(3)

1 module out of date using v2 of the Build Service, when the latest release (spec v2 component) needs to use v3:
Screenshot_2021-02-12 Origami Build Service(1)

an example error case (requesting a non-origami module "jquery"):
Screenshot_2021-02-12 Origami Build Service(2)


lib/routes/url-builder.js
```diff
@@ -0,0 +1,76 @@
...
+        if (item) {
+            item.current = true;
+        }
+        request.navigation = navigation;
+        next();
+    };
+}
+
+module.exports = app => {
+    // v2 home page
```
notlee, line 29:
```suggestion
```

lib/routes/url-updater.js
```diff
@@ -0,0 +1,75 @@
...
+        } catch (error) {
+            let errorMessage = error.message;
+            if (!(error instanceof UserError)) {
+                Raven.captureException(error);
+                console.error(error, JSON.stringify(error));
+                errorMessage = 'Sorry something went wrong whilst evaluating '
+                    + 'your Build Service URL against the latest component '
+                    + 'releases.';
+            }
+            return response.status(400).render('url-updater', {
```
JakeChampion, line 67: Question: Why do we handle the error response within this route handler instead of calling `next(error)`?

lib/url-updater/evaluate-modules.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @param {Array[]} - An array of arrays representing requested modules, e.g. [['o-example', 'v1.0.0'], 'o-test-component', '']
+ * @return {ModuleResult[]} - Details about the requested modules and how up to date they are.
+ */
+module.exports = async function evaluateModules(modules) {
+    const repoData = new RepoDataClient({});
+    const results = [];
+
+    for await (const module of modules) {
```
JakeChampion, line 26: According to the JSDoc `modules` is an array of arrays, which means we dont need `for await.. of` as we are not iterating over async iterable. We should change this for a `for.. of`.

```suggestion
    for (const module of modules) {
```

lib/url-updater/parse-build-service-url.js
```diff
@@ -0,0 +1,29 @@
...
+ * @throws {UserError} - Throws a user error if the provided build service is not valid or expected
+ */
+module.exports = function parseBuildServiceUrl(givenBuildServiceUrl) {
+    // Avoid XSS, error if the Build Service URL contains unexpected characters.
+    if (!givenBuildServiceUrl) {
+        throw new UserError('Could not parse your Build Service URL, looks like you didn\'t submit one to check yet?');
+    }
+    if (
+        typeof givenBuildServiceUrl !== 'string' ||
+        givenBuildServiceUrl.match(/[^\w\s-@^\.\*\,=:/?&\#\%]/)
```
JakeChampion, line 17: Can we have a comment explaining what this regex is meant to do, it would help when 'future us' comes to look at the code

lib/routes/url-updater.js
notlee: `next(error)` is generic and less helpful, this keeps the user on the update and lets them more easily correct any issue

lib/url-updater/parse-build-service-url.js
notlee:
```suggestion
        givenBuildServiceUrl.match(/[^\w\s-@^\.\*\,=:/?&\#\%]/) // match anything that is not a word (alphanumeric or underscore), space, dash, or other characters we expect in a Build Service URL
```

JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v357.2.0 :tada:

PR #448, opened 2021-02-15
Remove cache control from post requests `url-updater`

notlee:
We don't want to cache results without taking into account the post
data.

### DIFF
--- a/lib/routes/url-updater.js
+++ b/lib/routes/url-updater.js
@@ -34,7 +34,7 @@ module.exports = app => {
         });
     });
 
-    app.post('/url-updater/', addNavigationToRequest(), express.urlencoded({ extended: false }), express.text(), async (request, response) => {
+    app.post('/url-updater/', cacheControl({ maxAge: 0 }), addNavigationToRequest(), express.urlencoded({ extended: false }), express.text(), async (request, response) => {
         try {
             const url = parseBuildServiceUrl(request.body['build-service-url']);
             const brand = url.searchParams.get('brand');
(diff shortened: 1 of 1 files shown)


lib/routes/url-updater.js
```diff
@@ -34,7 +34,7 @@ module.exports = app => {
         });
     });
 
-    app.post('/url-updater/', cacheControl({ maxAge: '5m' }), addNavigationToRequest(), express.urlencoded({ extended: false }), express.text(), async (request, response) => {
+    app.post('/url-updater/', addNavigationToRequest(), express.urlencoded({ extended: false }), express.text(), async (request, response) => {
```
JakeChampion, line 37: Shall we explicitly state no caching of this resource?

To not cache in Fastly:
```
Cache-Control: private
```

To not cache in Fastly and the browser:
```
Cache-Control: private, no-store
```

notlee: Fastly doesn't cache post data by default, but yeah it's better to be explicit

JakeChampion: Do we want to cache it in the browser at all?

notlee:
```suggestion
        app.post('/url-updater/', cacheControl({ maxAge: 0 }), addNavigationToRequest(), express.urlencoded({ extended: false }), express.text(), async (request, response) => {
```

this won't set `private` but I'm happy to rely on default POST behaviour
https://github.com/Financial-Times/origami-service/blob/deaf7fedba9a61df712ab056d75c2d5fdaa158ed/lib/middleware/cache-control.js#L22

JakeChampion (approved):

Issue #455, opened 2021-02-22
v2: create a modules allow list


notlee:
v3 will only bundle components / demos which are published under `@financial-times`:
https://github.com/Financial-Times/origami-build-service/issues/453

v2 currently allows bundles for any package. Deprecate this feature with an allow list
of non-Origami modules which are already in use. Since v2 of the build service will be 
deprecated in favour of v3, this may be implemented with a hard-coded list of all allowed
modules.

notlee:
An allow list has been created but no errors are thrown yet, a warning is logged to Sentry instead:
https://sentry.io/organizations/financial-times/issues/2231501783/
When we know the allow list is comprehensive we can start throwing an error.

notlee:
The `shrinkwrap` param now uses the allow list:
https://github.com/Financial-Times/origami-build-service/pull/500
We can soon start confidently throw an error for non-allowed modules requested
with the `modules` param but will have to wait again before throwing for the `shrinkwrap`
param as it may require that additional modules are added to the allow list (dependents of 
the modules which are included with the `modules` param).

JakeChampion:
> We can soon confidently throw an error for non-allowed modules requested with the
> `modules` param but will have to wait again before throwing for the `shrinkwrap` param
> as it may require that additional modules are added to the allow list (dependents of the
> modules which are included with the `modules` param).

Since releasing https://github.com/Financial-Times/origami-build-service/pull/501, we now throw an error for non-allowed modules requested with the `modules` param.

We plan to do the same for the `shrinkwrap` param at some point.

@notlee after we do the same work for the `shrinkwrap` param, is that this issue completed?

notlee:
We're [already throwing an error for the shrinkwrap param](https://github.com/Financial-Times/origami-build-service/pull/501/files#diff-9782bf3e00d1e8d73e4a51d0a4ea35a62e32b4322ccf0ebc284b7d48eb4f4dc9R30), since we added every module dependency ever that's registered in repo data to the shrinkwrap allow list. But we continue logging a warning to Sentry as a precaution, so we can be reactive if there are any mistakes. I'd suggest we remove that logging and close this issue ~19th April, a month after the logs for the shrinkwrap param were added.

JakeChampion:
@github, remind us to review this issue on 19th of April please

PR #457, opened 2021-02-22
Report modules requested in v2 bundles which are not on an allow list

notlee:
When the allow list is comprehensive and covers any missed modules,
which should be allowed, the allow list can be enforced by throwing
an error for any disallowed module requested.

Since v2 will be deprecated for v3 of the Build Service soon a
static allow list should be okay.


origamiserviceuser:
:tada: This PR is included in version v357.3.0 :tada:

PR #464, opened 2021-03-02
configure project to use tabs for indenting

JakeChampion:
Currently the project did not have a consistent rule for indenting which lead to files being a mixture of tabs and spaces.

This pull-requests fixes this by making sure eslint will fail is the project has any lines of code which use spaces as the form of indentation

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,9 +6,10 @@ module.exports = {
 		'es6': true
 	},
 	'parserOptions': {
-        'ecmaVersion': 2020
-    },
+		'ecmaVersion': 2020
+	},
 	'rules': {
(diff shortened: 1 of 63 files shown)


.eslintrc.js
```diff
@@ -6,9 +6,10 @@ module.exports = {
 		'es6': true
 	},
 	'parserOptions': {
-        'ecmaVersion': 2020
-    },
+		'ecmaVersion': 2020
+	},
 	'rules': {
+		'indent': ['error', 'tab'],
```
JakeChampion, line 12: @notlee This was the only manual change.

All the rest was done by `eslint --fix`

notlee (approved):

Issue #454, opened 2021-02-22
v3: support non `o-` names


notlee:
I happened to spot an assumption that component names begin with `o-` in the codebase:
https://github.com/Financial-Times/origami-build-service/blob/8932e483fce442423ffa19dd0485174e6b21d48b/lib/middleware/v3/createEntryFileSass.js#L24

[That's not necessarily true](https://origami.ft.com/spec/v1/components/#naming-conventions), `o-` is encouraged but not enforced, e.g. g-audio.

Task: add v3 tests for CSS bundles, JS bundles, and demos of Origami compatible components with a name not prefixed with `o-`.

JakeChampion:
> Task: add v3 tests for CSS bundles, JS bundles, and demos of Origami compatible components with a name not prefixed with `o-`.

Good idea! This requires making a new test origami component. We could publish o-test-component under two names? `test-component` and `o-test-component`... Or we could just start publishing it as `test-component`?

What do we think @Financial-Times/origami-core?

PR #473, opened 2021-03-05
Make build-service v3 css endpoint work for non o- prefixed components

JakeChampion:
This resolves https://github.com/Financial-Times/origami-build-service/issues/454

### DIFF
--- a/lib/middleware/v3/createEntryFileSass.js
+++ b/lib/middleware/v3/createEntryFileSass.js
@@ -21,7 +21,7 @@ async function createEntryFileSass(installationDirectory, components, brand, sys
 	for (const name of componentNames) {
 		const importComponent = `@import "${name}";`;
 		let include = '';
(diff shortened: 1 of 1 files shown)


lib/middleware/v3/createEntryFileSass.js
```diff
@@ -21,7 +21,7 @@ async function createEntryFileSass(installationDirectory, components, brand, sys
 	for (const name of componentNames) {
 		const importComponent = `@import "${name}";`;
 		let include = '';
-		if (name.startsWith('@financial-times/o-')) {
+		if (name.startsWith('@financial-times/')) {
```
notlee, line 24: 👍 Would it be worth adding a test for this, including one to check that a helpful error is returned for non-ft modules?

notlee: I see you've done that in another PR already :)

notlee (approved):

PR #477, opened 2021-03-09
speed up integration tests by using the asserting the headers and body of a single response

JakeChampion:
our previous approach was to make a new request/response for each assertion, even if the request was for the exact same resource we had just requested.

Running the integration tests against a local application:
- Without this change: 4 minutes 5 seconds
- With this change: 1 minute 30 seconds.

Running the integration tests against the heroku application:
- Without this change: 8 minutes 43 seconds
- With this change: 3 minutes 26 seconds


chee (approved):
very cool

notlee (approved):

PR #479, opened 2021-03-11
return an error response when attempting to build a demo whose html is invalid.

JakeChampion:
If a component is published and it's demo contains incorrect html we have decided that the service should respond with an error message stating the html is incorrect/invalid because we know some of our users copy/paste the demo html and we don’t want to be responsible for copy-pasted html errors within our users' products.


lib/middleware/v3/outputDemo.js
```diff
@@ -37,7 +37,13 @@ const rimrafOptions = {
 };
 
 const extractMinimalHtml = (fullHtml) => {
-	const $ = cheerio.load(fullHtml);
+	let formattedHTML;
+	try {
+		formattedHTML = prettier.format(fullHtml, { parser: 'html' });
+	} catch (error) {
+		throw new ComponentError('The HTML in the demo contains syntax errors.');
```
JakeChampion, line 44: @Financial-Times/origami-core, any suggestions on how we could improve this error message for when the demo html is syntactically invalid?

chee: Does the error itself have information about where the parsing errors took place? can we print those to give them idea of where to look?

JakeChampion: Yes, it has it inside a very big string which contains the entire html document. I'll make it return that 👍

JakeChampion: Done in [`f34696f` (#479)](https://github.com/Financial-Times/origami-build-service/pull/479/commits/f34696fd1f46e895c73f74ba6fefec65ca1c8b71)

chee (approved):

PR #480, opened 2021-03-11
Ready for review: Handle slow response generation on the v3 endpoints

JakeChampion:
This pull request makes all the v3 endpoints have an in-memory cache which stores the results for 2 minutes.

The reason to add this cache is because sometimes a response will take more than 30 seconds to generate (usually because the `npm install` is taking a while to finish) and Heroku's router will close any connections which have not returned a response within 30 seconds.

I've also added logic to detect when a response is going to take longer than 25 seconds and respond with a redirect back to the same endpoint with the same parameters. This is to make the browser remake the same request to the server, which will enable the server to locate the response in the cache which is already being generated based on the previous request instead of generating a new response from scratch.

### DIFF
--- a/lib/middleware/v3/createCssBundle.js
+++ b/lib/middleware/v3/createCssBundle.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const querystring = require('querystring');
 const fs = require('fs').promises;
 const path = require('path');
 const Raven = require('raven');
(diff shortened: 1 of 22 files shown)


JakeChampion:
The only way I can think to write an integration test for this scenario will involve the test knowing the internal implementation of origami-build-service...

I thought we could use a HTTP interceptor and slow down the npm registry HTTP calls but we can't do that as the application process does not make those HTTP calls itself, the application spawns a child-process which execute the npm command-line application  and that child-process makes the npm HTTP calls.

If the test is allowed to have knowledge of how the application installs dependencies, the test could intercept the child-process communication and add an artificial delay to the response from the child-process, thereby triggering the redirection logic that we are adding in this pull-request.

lib/middleware/v3/createCssBundle.js
```diff
@@ -17,76 +14,93 @@ const { installDependencies } = require('./installDependencies');
...
-		response.startTime('createPackageJsonFile');
-		await createPackageJsonFile(bundleLocation, components);
-		response.endTime('createPackageJsonFile');
+	return bundle;
+}
 
-		response.startTime('installDependencies');
-		await installDependencies(bundleLocation, request.app.ft.options.npmRegistryURL);
-		response.endTime('installDependencies');
+const twoMinutes = 120 * 1000;
```
chee, line 48: that's true

chee (approved):

PR #482, opened 2021-03-15
Add "o-three-sixty" to the v2 bundle allow list.

notlee:
It was missed when searching Splunk for modules requested within
the last 2 months, but has had a request recently. This may be
because success logs are sampled.

"o-three-sixty":
https://github.com/ftlabs/o-three-sixty

Related issue:

### DIFF
--- a/lib/data/v2-module-allow-list.json
+++ b/lib/data/v2-module-allow-list.json
@@ -12,6 +12,7 @@
     "n-swg",
     "sass-mq",
     "superstore-sync",
+    "o-three-sixty",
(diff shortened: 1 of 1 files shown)


chee (approved):
neat

notlee:
>This may be because success logs are sampled.

Actually, the Splunk query used to compile this list ignored `o-` modules with the assumption they were Origami components which would be included in this list already. "o-three-sixty" isn't in repo data and doesn't show in the registry anymore.

PR #485, opened 2021-03-15
update v2 test to expect the "2.0" version string

chee:
Does this work?


JakeChampion:
> Does this work?

This will work, we will need to release new versions of o-test-component which have the correct origami version defined though as the current integration tests are using a version of o-test-component that will not pass this updated version of isOrigamiV2Component

chee:
it is ready

notlee (approved):
Excellent :) I love the patches have stayed the same in the updated test component

PR #491, opened 2021-03-16
Add autoprefixer for css and use same browser config as origami-build-tools

JakeChampion:
This is required to ensure the css returned by the build-service will work correctly for the browsers we are attempting to support. I've used the same configuration for browser support as origami-build-tools because that configuration accounts for the browsers listed in the FT's official browser support policy.


JakeChampion:
Thinking out loud:

We could have an integration test for this by releasing a version of o-test-component that uses properties we know should get prefixed by our auto-prefixer configuration and assert that the return css bundle contains those prefixed properties.

notlee (approved):

notlee:
>We could have an integration test for this by releasing a version of o-test-component that uses properties we know should get prefixed by our auto-prefixer configuration and assert that the returned css bundle contains those prefixed properties.

That would be good but also I think this is a sensible test to cover it :)

PR #496, opened 2021-03-19
Fix OBS v2 endpoints given component dependencies with invalid origam…

notlee:
…i.json

`/v2/files/o-header@2.5.13/img/ft-logo.png` currently fails as an install
of `o-viewport`, a dependency, has an invalid `origami.json` that cannot
be parsed.

Instead, when it is not possible to check if a component follows the v1 spec
due to `origami.json` issues assume that it does so not to break existing
requests.

Related PR: https://github.com/Financial-Times/origami-build-service/pull/438
Fixes: https://sentry.io/organizations/financial-times/issues/2254274035/?project=52558&query=is%3Aunresolved

### DIFF
--- a/lib/cssbundler.js
+++ b/lib/cssbundler.js
@@ -100,11 +100,17 @@ CssBundler.prototype = {
 		const components = yield installation.listAllOrigamiComponents();
 
 		for (const [name, properties] of Object.entries(components)) {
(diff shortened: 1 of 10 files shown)


JakeChampion (approved):
Good work finding these regressions and fixing them 👍

PR #500, opened 2021-03-22
Report non-allowed shrink-wrapped modules requested in v2 bundles

notlee:
The shrinkwrap parameter allows code from any repository to be
bundled, skipping the allow list check. Requested modules
in both parameters should be checked.

Since v2 will be deprecated for v3 of the Build Service soon a
static allow list should be okay.

Related PR: https://github.com/Financial-Times/origami-build-service/commit/ed17062b15f1e45cea08252d4a6cf3489e136ff4
Issue: https://github.com/Financial-Times/origami-build-service/issues/455

### DIFF
--- a/lib/middleware/reportDisallowedModules.js
+++ b/lib/middleware/reportDisallowedModules.js
@@ -12,19 +12,40 @@ module.exports = function () {
      */
 	return (request, response, next) => {
 
-		// Used by bundle endpoints
-		for (const [name, version] of parseModulesParameter(request.query.modules)) {
-			if (!allowedModules.includes(name)) {
-				Raven.captureMessage('A module was requested via v2 of the Build Service which is not on the allowed list of modules.', {
-					level: 'warning',
-					tags: {
-						disallowed_module: true,
-						module: name,
-						version
-					},
-				});
+		// modules param: Used by bundle endpoints.
+		if(request.query.modules) {
+			for (const [name, version] of parseModulesParameter(request.query.modules)) {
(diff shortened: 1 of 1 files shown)


lib/middleware/reportDisallowedModules.js
```diff
@@ -12,19 +12,41 @@ module.exports = function () {
...
+				}
 			}
+
+		}
+
+		// shrinkwrap param: Used by bundle endpoints.
+		if(request.query.shrinkwrap) {
+			for (const [name, version] of parseModulesParameter(request.query.shrinkwrap)) {
+				if (!allowedModules.includes(name)) {
+					throw new Error('A module was requested in the shrinkwrap parameter, via v2 of the Build Service, which is not on the allowed list of modules.');
```
JakeChampion, line 36: I think we should remove this error for now, instead we can carry on allowing the requests and capturing the issues in Sentry.

Later on when we have more data about how this parameter is being used, we can update the allow list and then add the code to start throwing the errors.

If we use the approach I've outlined, then we would know how many requests we will no longer be supporting, which helps us figure out the impact of this change

notlee: This was totally a mistake! Left over from testing, that error is not supported to be there. Thanks for spotting.

Yes we can soon throw for the modules param but will have to wait again for the shrinkwrap param, as it may include extra modules. Or we can check every dependency of the allow list and ensure it is included (the shrinkwrap param should include only dependencies of the modules param).

JakeChampion (approved):

PR #510, opened 2021-04-19
Do not report to Sentry module/shrinkwrap allow list misses

notlee:
We are more confident the allow list for the modules/shrinkwrap
param covers existing requests.


JakeChampion (approved):
🏄

PR #515, opened 2021-04-27
allow bundle requests to contain components which do not have js and/or sass

JakeChampion:
This is because origami component demos use a single list of dependencies to generate the script and link tags in their html, which means they regularly have a script and/or link tag which is asking for components which don't export js/sass

This wasn't noticed until now because I had not updated the origami-build-tools v11 to use the new version of the build service.

### DIFF
--- a/lib/middleware/v3/createEntryFileJavaScript.js
+++ b/lib/middleware/v3/createEntryFileJavaScript.js
@@ -1,8 +1,15 @@
 'use strict';
 
 const path = require('path');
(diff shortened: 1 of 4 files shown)


notlee (commented):
Nice catch. That could have been an issue for users too, I never check I always add a component to both js/css requests.

Looks good to me but how about adding a specific test case? It's a nice way to document these kind of features.

notlee (approved):

Issue #400, opened 2020-12-01
Use origami-repo-data to fetch list of components to add to the npm cache upon server booting up


JakeChampion:
Currently the project is using a fixed this.This probably shouldn't exist as a fixed list.
It should use an origami-repo-data api call to get the list similar to origami-component-converter and fallback to a fixed list if the origami-repo-data api call fails.

We can reuse some code in origami component converter which does a similar thing --
https://github.com/Financial-Times/origami-component-converter/blob/10990dac98e980594a329f0cd68537b0e51a6560/lib/npm.js#L8-L17

_Originally posted by @JakeChampion in https://github.com/Financial-Times/origami-build-service/pull/391#discussion_r532691930_

chee:
If we're going to be doing an Origami repo data call, then we can also perhaps build this web page too: https://github.com/Financial-Times/origami-build-service/issues/39

what do you think?

JakeChampion:
> If we're going to be doing an Origami repo data call, then we can also perhaps build this web page too: #39
> 
> what do you think?

That's a great idea 👍

PR #516, opened 2021-04-29
Remove the code to cache a list of components on application boot

JakeChampion:
This approach only helps the first request to the application server which includes a component at the latest version and that request isn't cached in the cdn.

We will more than likely have a cached version in the cdn, if we don't it is more than likely that a previous request has already made it to the application server which placed the components into the npm cache automatically.

Closes https://github.com/Financial-Times/origami-build-service/issues/400


chee (approved):
honk!

PR #517, opened 2021-04-29
install the requested components as peer-dependencies

JakeChampion:
currently the service is installing the components as dependencies

it is meant to install them as peer-dependencies to ensure that the dependency tree for origami components is flat


chee (approved):
honk!

PR #532, opened 2021-06-29
Improve URL Updater to help with OBS v2 to v3 migrations.

notlee:
**I would describe this as scrappy:** Test coverage is pretty weak. It's a development tool and most aspects of this code should be short lived, 1yr for v2 migrations, so I'm not too concerned about that.

- Help users migrate to the latest Bower release, separate
to migrating from OBS v2 to v3.
- Include `o-autoinit` in the `components` param if missing.

1. Users are directed to upgrade to the latest Bower release of components before moving to OBS v3.
Screenshot 2021-07-02 at 11-31-37 Origami Build Service
2. No code changes should be required for OBS users completing the npm migration. There's a call to contact us though if they have any issues.
Screenshot 2021-07-02 at 11-31-53 Origami Build Service
3. If a user is working on the npm migration but there has been another npm-only release since, then the user is prompted to follow the migration guide. This screenshot links to the migration guide for o-test-component v3, since v2 was the first npm-only release.
Screenshot 2021-07-02 at 11-33-57 Origami Build Service
4. We're all done 
Screenshot 2021-07-02 at 11-32-12 Origami Build Service


lib/url-updater/update-url-for-results.js
```diff
@@ -1,27 +1,43 @@
...
 		url.toString().replace(
 			/(?<=service\/build\/v)([1-2])(?=\/)/,
 			minimumBuildServiceVersion
 		));
 
-
 	const componentParam = minimumBuildServiceVersion === 2 ? 'modules' : 'components';
 	const unsupportedComponentParam = minimumBuildServiceVersion === 2 ? 'components' : 'modules';
+
+	const componentParamValue = results.map(c => `${c.name}@${hasComponentsBehindTheLatestBowerMajor ? (c.belowLatestBowerRelease ? `^${c.lastBowerMajor}` : c.requestedVersion) : `^${c.latestVersion}`}`);
```
JakeChampion, line 27: non-blocking suggestion: I had a hard time reading the nested template literals and ternaries here. To understand what was going on, I changed it to a nested if statement for setting only the version part of the final string.
Does this look correct @notlee?
```suggestion
	const componentParamValue = results.map(c => {
		let version = `^${c.latestVersion}`;
		if (hasComponentsBehindTheLatestBowerMajor) {
			if (c.belowLatestBowerRelease) {
				version = `^${c.lastBowerMajor}`;
			} else {
				version = c.requestedVersion;
			}
		}
		return `${c.name}@${version}`;
	};
```

lib/url-updater/update-url-for-results.js
```diff
@@ -1,27 +1,43 @@
...
+	const includesAutoInitComponent = results.find(({name}) => name === 'o-autoinit');
+	if (minimumBuildServiceVersion === 3 && !includesAutoInitComponent) {
```
JakeChampion, line 29: This is missing some conditions, if the original build service request was to v2, then there is a query parameter named `autoinit` which we need to look at, if it is set to `0` then we should not add `o-autoinit`

The [documentation](https://www.ft.com/__origami/service/build/v2/docs/api) for the query parameter `autoinit`:
> (Optional) If absent, or present and set to a truthy value, the bundle will include code to dispatch the o.DOMContentLoaded and o.load events when their browser-native counterparts occur. If set to 0, no auto-initialisation code will be included.

views/partials/url-updater/v2-migration.html
```diff
@@ -0,0 +1,39 @@
+<p>Your Origami Build Service URL includes outdated components{{#if hasFurtherNpmOnlyMajorReleases}}, some are more than one major version{{else}} which are a major version{{/if}} behind the latest release:
+
+<ol>
+    <li>
+        <p>To upgrade, start by using the below Origami Build Service URL which requests the same components but with updated version numbers, and includes the latest version of <a href="https://registry.origami.ft.com/components/o-autoinit@2.0.7/readme?brand=master">o-autoinit</a>:</p>
```
JakeChampion, line 5: The section about o-autoinit will need to be behind a boolean as the build service request may not include o-autoinit. This can happen if the original vs build service request has set the query parameter `autoinit` to `0`

views/partials/url-updater/v3-major.html
```diff
@@ -0,0 +1,24 @@
+<p>Your Origami Build Service URL includes outdated components, some are one or more major versions behind the latest release:
+
+<ol>
+    <li>
+        <p>To upgrade, start by using the below Origami Build Service URL which requests the same components but with updated version numbers, and ensures <a href="https://registry.origami.ft.com/components/o-autoinit@2.0.7/readme?brand=master">o-autoinit</a> is included:</p>
```
JakeChampion, line 5: Same comment here as https://github.com/Financial-Times/origami-build-service/pull/532/files#r662937100

views/partials/url-updater/v3-minor.html
```diff
@@ -0,0 +1,11 @@
+<p>
+    To upgrade, use the below Origami Build Service URL which requests the same components but with updated version numbers, and ensures <a href="https://registry.origami.ft.com/components/o-autoinit@2.0.7/readme?brand=master">o-autoinit</a> is included.
```
JakeChampion, line 2: Same comment here as https://github.com/Financial-Times/origami-build-service/pull/532/files#r662937100

JakeChampion (changes requested):
Nice work on this @notlee

I've left some comments about the o-autoinit part which we should address before merging, everything else is great 👍

JakeChampion (approved):
nice!

PR #536, opened 2021-07-20
Link to version 3 of o-autoinit

JakeChampion:
We currently link to the older version of o-autoinit on the url migration page - we should be linking to version 3, which is the version that is npm-only and works with build service v3


chee (approved):

origamiserviceuser:
:tada: This PR is included in version v362.1.3 :tada:

PR #537, opened 2021-07-23
bugfix: make the v3 exported Origami variable to not have the `.default` property on the components

JakeChampion:
We want to match the behavior of the v2 export Origami variable which is to have the components' code be available directly from the variable. E.G.`let Table = window.Origami['o-table'];` should work in v2 and in v3.

Currently it does not do this due to changing the bundler being used to ESBuild, which puts the exports component code on a `.default` property.

This is a breaking change, but we should try and do it if possible because right now it complicates the migration process more than we want and is behavior we would like to not have.

We should check the logs to see if anyone is using v3 in production and see if this behavior is already being relied upon. I know origami services moved to the v3 api but they do not rely on the exported Origami variable, hopefully this is the same for anything else which is using v3 right now 🤞


chee (approved):

notlee:
This should now allow the component api to be accessed with and without referencing the `.default` property as @chee suggested. Including if a `callback` function has been given. We can follow up with our users and make sure they are not come to rely on `.default` and then remove it. Here is a list of potentially affected system codes:

- bizzabo-ft-live
- origami-registry-ui
- ig-content
- biz-ops-admin
- origami
- build-service
- origami-repo-data
- deafblind-media
- staff-ft-com-redirect-page

test/unit/lib/middleware/v3/createJavaScriptBundle.test.js
```diff
@@ -85,9 +85,9 @@ describe('createJavaScriptBundle', function () {
 
 			const bundle = response._getData();
 
-			proclaim.deepStrictEqual(
+			proclaim.include(
```
JakeChampion, line 88: The version of o-test-component being used here doesn't export anything from it's main.js
I've started a pull-request to add a version of o-test-component which does export from main.js, that way we can be certain this is working as we expect it to

PR: https://github.com/Financial-Times/o-test-component/pull/187

JakeChampion: If we use v3.1.0 of o-test-component, we could then add a test assertion that shows `context.Origami["o-test-component"] === context.Origami["o-test-component"].default`

JakeChampion (commented):
I can't hit "Approve" as I opened the pull-request --

I do approve of this pull-request 👍

notlee (approved):

PR #544, opened 2021-08-05
Set the caching for post responses to /url-updater to be private

### LINKED ISSUE #543: Investigate the caching issue on the url updater page
## What
Occasionally when submitting a url to the url-updater page, the page loads a stale response which is incorrect

## Details
- Paste an outdated url into the url-updater form and submit the form -- keep doing this until you get a stale response... it is very inconsistent

JakeChampion:
This resolves #543.

This informs Fastly that the response should not be reused for other requests which are currently waiting for a response (A feature known as request collasping).

I  ran 1000 post requests against the production version and 1881 of those returned a cached copy of the wrong page 🙈 

I then ran 1000 post requests to the url-updater page of this branch via Fastly and none of them are returning a cached copy of the wrong page 🎉

### DIFF
--- a/lib/routes/url-updater.js
+++ b/lib/routes/url-updater.js
@@ -19,8 +19,9 @@ module.exports = app => {
 		});
 	});
 
-	app.post('/url-updater/', cacheControl({ maxAge: 0 }), addNavigationToRequest(), express.urlencoded({ extended: false }), express.text(), async (request, response) => {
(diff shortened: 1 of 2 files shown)


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v362.1.5 :tada:

PR #545, opened 2021-08-06
configure npm to use a cache directory which is stored within the app

JakeChampion:
this enables us to have commit cached files into the project and have them always be in the npm cache for the build-service to use

previously it would use the global npm cache folder which gets removed each time the app is deployed to heroku and every 24 hours on heroku

If this looks to speed up the apis considerably then we should look into using git lfs for storing the cache instead of committing the files to the repo directly - we also would need to a way to add the new versions into the cache


JakeChampion:
@Financial-Times/origami-core  This commit contains all the changed code which should be reviewed -- https://github.com/Financial-Times/origami-build-service/commit/74fff67017bab706cc3e7c4e64251d890660ee69

notlee (approved):
non-blocking suggestion: It's quite hard to tell what the difference is in the try and catch block without the context of this PR. It'd be great if we could make that clearer, maybe a code comment or by wrapping the npm command in a well named function (similar in `installDependencies.js `)

chee (approved):

JakeChampion:
> non-blocking suggestion: It's quite hard to tell what the difference is in the try and catch block without the context of this PR. It'd be great if we could make that clearer, maybe a code comment or by wrapping the npm command in a well named function (similar in `installDependencies.js `)

I've extracted them out into separate functions now 👍  [`28abe7d` (#545)](https://github.com/Financial-Times/origami-build-service/pull/545/commits/28abe7d3c30c563f362fec67424ddb4adef1e10c)

notlee (approved):

PR #556, opened 2021-11-03
feat: rename "master" brand to "core"

notlee:
Map "core" to "master" to continue to support previous versions of components. Components will drop "master" support in a future major version. We will need to release a new major version of the Build Service which does not map "core" to "master" to support future major versions of components.


notlee:
@Financial-Times/origami-core this is ready for review now

JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v363.1.0 :tada:

PR #557, opened 2021-11-08
fallback to running npm without any --offline or --prefer-offline flags

JakeChampion:
it seems the documentation for --prefer-offline is not aligned with the implementation

We've seen multiple times now that when a new version of a component has been published, the build service cannot locate that version when running `npm install --prefer--offline` or `npm pack --prefer-offline` but it can find the version when run without that flag.

This commit introduces a new set of functions to use when installing a dependency. Now the installation flow goes like so:
- Try installing using the cached copy and do not attempt to use the network at all
- If that failed, try installing using the cached copy and use the network as a last resort
- If that failed, use the network first and if successful then update the cached copies

### DIFF
--- a/lib/middleware/v3/downloadAndUnpackPackage.js
+++ b/lib/middleware/v3/downloadAndUnpackPackage.js
@@ -55,6 +55,27 @@ async function downloadFromCacheWithNetworkFallback(location, name, version, reg
 }
 
 /**
+ *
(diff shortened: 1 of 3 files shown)


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v363.2.0 :tada:

PR #558, opened 2021-12-22
When building a demo, install all dependencies instead of only production dependencies

JakeChampion:
This will enable origami component's to define their demo dependencies in the package.json.devDependencies field instead of origami.json.

This would make keeping demo dependencies up-to-date simpler to achieve and would also mean that demo dependencies would be linked in the origami monorepo - E.G. When we update the design to o-buttons, we would have all the demos for all component already using the new design without having to publish anything to npm -- something which we currently can't do as the demo dependencies are downloaded from the build-service, which means they must first be published to npm.


JakeChampion:
We'd also be able to user non origami packages within only the demos, that is not possible with the current workflow listing demo dependencies in origami.json and turn that into a build-service bundle

chee (approved):
i approve

PR #576, opened 2022-03-29
attach all route handlers under the path prefix /__origami/service/build/

JakeChampion:
this is because the production origami build service is served under www.ft.com/__origami/service/build/ and having the application server respond to this path will simplify our cdn configuration

By making this change we can remove these parts of the cdn configuration:
https://github.com/Financial-Times/ft.com-cdn/blob/537963bad7f1ca121177855cb574df6f1c59150d/src/vcl/origami-build-service.vcl#L28
and
https://github.com/Financial-Times/ft.com-cdn/blob/537963bad7f1ca121177855cb574df6f1c59150d/src/vcl/origami-build-service.vcl#L36-L38
and
https://github.com/Financial-Times/ft.com-cdn/blob/537963bad7f1ca121177855cb574df6f1c59150d/src/vcl/origami-build-service.vcl#L87-L91

The reason we have added new routes instead of replacing the existing routes is
because when this is deployed it will need to respond to the old routes whilst
we make the changes to the CDN to remove the vcl which remove the path prefix
/__origami/service/build/ before making the request to our application servers


JakeChampion:
> It was super nice that you deleted [basePath](https://github.com/Financial-Times/origami-image-service/pull/787/files#diff-ba85e65e5381efabd2a2e48851c18adee2030e7d64e82265f3a041ebea630a7cL32) in the similar image service PR, is that something we could do here too?

Yep, I have done that locally but didn't finish it yet 👍 I'll mark this PR as a draft until that is done

notlee (approved):
🙌

origamiserviceuser:
:tada: This PR is included in version v364.1.0 :tada:

PR #577, opened 2022-03-30
fix: filter o-fonts-assers from v bundle requests

notlee:
o-fonts-assets should not be required directly by JS or CSS bundles and can cause an error when Github auth fails:
this is currently happening due to our bot account origamiserviceuser losing Github access. As future access will
be through Okta I don't expect this to be restored quickly.

o-fonts-assets is the only component I know of which is private. However other build service requests could in theory
request private repos which are not Origami components.
https://github.com/Financial-Times/origami-build-service/blob/e61068dd6b0dfa2a6159fb988a4578f9df0b549b/index.js#L35

### DIFF
--- a/lib/middleware/removeFontsAssetsFromModulesParameter.js
+++ b/lib/middleware/removeFontsAssetsFromModulesParameter.js
@@ -1,0 +1,35 @@
+'use strict';
+
+const parseModulesParameter = require('../utils/parseModulesParameter');
+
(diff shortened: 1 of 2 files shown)


notlee:
Tests aren't completing here because of the same Github auth issue that requires this PR 😅No one else is around right now so I'm going to release this 😅 😅

PR #586, opened 2022-07-07
backstage: update archive metric logging

notlee:
archive.miss = a valid request which would be handled was not by the archive; archive.fallback = some request was not handled by the archive but may not have been handled anyway


notlee:
Please review this Origami team

notlee:
Looks good to me, if tests pass

notlee:
Thanks Origami team

PR #588, opened 2022-07-08
feat: update v2 archive to handle cache busted urls

notlee:
Some sites are included Build Service v2 requests with a cache
busting query param. I’ve found at least 220 affected sites such as https://smg2022.ft.com and https://globalmacrooutlook2022.live.ft.com, powered by bizzabo. These sites update the cache bust on every request(!?). Instead of updating hundreds of these pages and missing others, the archive has been updated to strip all but expected query params.

Expected query params is set based on a unique list of parameter names found in all currently archived urls, minus the made up cache busting param names.

### DIFF
--- a/lib/middleware/archive.js
+++ b/lib/middleware/archive.js
@@ -1,31 +1,8 @@
 'use strict';
 
-const crypto = require('crypto');
 const Raven = require('raven');
 const axios = require('axios').default;
-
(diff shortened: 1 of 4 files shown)


origamiserviceuser:
:tada: This PR is included in version v364.2.4 :tada:

PR #592, opened 2022-07-25
feat: Remove OBS v2 ARCHIVE env / toggle. Always use the archive.

notlee:


akomiqaia (approved):

notlee:
Huh there are quite a few test error here. Interestingly I didn't see any when running locally. Will investigate before merging 🤔

notlee:
Tests for bundle and file endpoints are returning an error because this PR make all endpoints use only the archive. Removing a code path to previous behaviour. To make this easier to review I've decided to close this PR and change the archive to be always on route by route. https://github.com/Financial-Times/origami-build-service/pull/593

PR #594, opened 2022-07-25
feat!: Always use archive for OBS v2 files.

notlee:
Unlike V2, the OBS v1 `files` endpoint now returns 410, gone. v1 was deprecated a long time ago. We are getting no requests for modules, looking at our metrics, and so are very unlikely to be serving files via the v1 endpoint either.


origamiserviceuser:
:tada: This PR is included in version v365.2.0 :tada:

PR #596, opened 2022-07-26
backstage: Remove all routes which do not begin with '__/origami/...'

notlee:
Our CDN has been updated and these are no longer needed.
https://github.com/Financial-Times/origami-build-service/pull/576

This commit also updates OBS v1 to redirect to v3 rather than v2.


lib/routes/v3/bundles.js
```diff
@@ -4,13 +4,11 @@ const {createJavaScriptBundle} = require('../../middleware/v3/createJavaScriptBu
 const {createCssBundle} = require('../../middleware/v3/createCssBundle');
 module.exports = app => {
 	app.get([
-		'/v3/bundles/js/',
```
akomiqaia, line 7: I am not sure but did you intend to remove v3 url?

akomiqaia: I guess you did since this kind of changes are everywhere

akomiqaia (approved):

lib/routes/v3/bundles.js
notlee: Yeah, our CDN has been updated so these are no longer needed. The app will responsed to `'/__origami/service/build/v3/bundles/js/` still.
https://github.com/Financial-Times/origami-build-service/pull/576

origamiserviceuser:
:tada: This PR is included in version v366.1.0 :tada:

PR #599, opened 2022-07-27
Update Surrogate-Key headers used for purging cache.

notlee:
The [Surrogate-Key header allows us to purge](https://docs.fastly.com/en/guides/purging-with-surrogate-keys) a batch of urls
from our CDN.

Currently the Origami Build Service v2 Surrogate-Key header
is set by our cdn, Fastly. Here's the VCL used to configure that:
https://github.com/Financial-Times/ft.com-cdn/blob/da01b97de677f74606f7ed76533905255c536cdd/src/vcl/origami-build-service.vcl#L34

Some problems:
- VCL is less understood at the FT. This is happening outside the
Origami Build Service codebase. It's harder to reason about.
- The Fastly service is for the whole of ft.com. "website" as a surrogate
key could easily clash.
- There are no Surrogate-Key headers set for v3 bundles, fonts, or
demos, making it difficult to purge these categories of pages if
we need to.

This PR:
- Moves existing Surrogate-Key headers to the Origami Build Service
codebase, so we can remove from ft.com-cdn and simplify our VCL.
- Namespaces Surrogate-Key headers `origami-build-service-*`
- Adds Surrogate-Key headers to v3 responses.
- Updates the readme.


akomiqaia (approved):

origamiserviceuser:
:tada: This PR is included in version v366.3.0 :tada:

PR #613, opened 2023-06-19
[Do not merge]  OR-230/return http 410 on deprecated fonts

frshwtr:
Deprecate `Miller` and `Benton` fonts on v2 and v3 build service requests.

The following requests, where `{font}` is either `Miller` or `Benton` variations:
```
GET /__origami/service/build/v3/fonts?font_name={font}...
GET /__origami/service/build/v2/file/o-fonts-assets@1.23.0/{font}
```

will no longer return the desired font, and instead respond with `410 Gone`


frshwtr:
Hold the presses, something is wrong with my prettier setup locally. Reviewable but please don't merge just yet.

lib/middleware/archive.js
```diff
@@ -3,11 +3,21 @@
...
 
 module.exports = function (app) {
 	const {metrics} = app.ft;
 
 	return async function outputBundle(req, res, next) {
+		if(req.originalUrl.includes('/files/o-fonts-assets')) {
+			const fontFileName = req.originalUrl.replace(/^\/__origami\/service\/build\/v2\/files\/o-fonts-assets@.*\//, '');
+
+			if(isFontDeprecated(fontFileName)) {
+				const obsError = new Error(`Origami Build Service returned an error: The font '${fontFileName}' is deprecated.`);
```
notlee, line 16: This error message doesn't provide much context, so a user expecting to get a font wouldn't necessarily know what to do. It's not really an error, being intentional. Maybe something like this?

"The font "${fontFileName}" is deprecated, we no longer have a license to use it. Please contact the Origami Team for advice on which font to use instead."

lib/middleware/v3/outputFont.js
```diff
@@ -94,6 +92,20 @@ const outputFont = async (request, response) => {
...
+		if (isFontDeprecated(fontFileName)) {
+			// We are using a content-type of text/plain to ensure that the browser/user-agent
+			// which recieves the response does not try and execute it.
+			// This protects our error responses from any form of cross-site-scripting (XSS) attack.
+			response.setHeader('Content-Type', 'text/plain; charset=UTF-8');
+			response.setHeader('X-Content-Type-Options', 'nosniff');
+			response.setHeader('Cache-Control', 'max-age=0, must-revalidate, no-cache, no-store');
+			response.status(410);
+
+			response.send(JSON.stringify(`Origami Build Service returned an error: The font '${fontFileName}' is deprecated.`));
```
notlee, line 105: As previous

lib/middleware/v3/outputFont.js
```diff
@@ -94,6 +92,20 @@ const outputFont = async (request, response) => {
...
 		const fontFileName = parseFontNameParameter(request.query.font_name);
 		const fontFormat = parseFontFormatParameter(request.query.font_format);
+
+		if (isFontDeprecated(fontFileName)) {
+			// We are using a content-type of text/plain to ensure that the browser/user-agent
+			// which recieves the response does not try and execute it.
+			// This protects our error responses from any form of cross-site-scripting (XSS) attack.
+			response.setHeader('Content-Type', 'text/plain; charset=UTF-8');
+			response.setHeader('X-Content-Type-Options', 'nosniff');
+			response.setHeader('Cache-Control', 'max-age=0, must-revalidate, no-cache, no-store');
```
notlee, line 102: I think it's right to be explicit here with the cache control header, to cache the 410 explicitly in a way which may vary from font files. However, perhaps we should cache this response for a short while? It's unlikely to change but we will want to have the option to return a font in a matter of minutes if we change our mind

notlee: Minutes, at least initially, we will want to have the option to roll back if we change our mind

notlee (approved):

notlee:
Do not merge until Wednesday 21st June

PR #620, opened 2023-11-29
fix: Increase cache duration for archived assets and demos

notlee:
Archived assets are unlikely to change, and neither are demos. I have introduced a more granular surrogate key so we can clear the cache for specific component demos if needed.

Screenshot 2023-11-29 at 09 33 32


akomiqaia (approved):

origamiserviceuser:
:tada: This PR is included in version v366.4.2 :tada:

PR #626, opened 2024-04-04
Remove url updater.

notlee:
This was created to support migrations away from the Bower package manager. It supported users to upgrade components requested via the Build Service, sometimes by many major versions.
https://origami.ft.com/blog/2021/01/18/deprecating-bower-and-origami-via-npm/

The Bower migration was completed many years ago. Though some sites sill use an old archive, it is unlikely they will complete the migration. Therefore the value of the url updater is limited. More importantly:

1. It will not support users to upgrade to new "o3" components, which will replace existing components and support a rollout side-by-side.
2. It is broken. We turned off origami-repo-data which it relied on.


ob6160 (approved):
👋

PR #628, opened 2024-04-26
[EI-4701] -  Use regex to validate system codes instead of making HTTP calls to biz-ops-system-codes

kiyaGu:
## Why
We had a discussion with the Origami team (@notlee) regarding the decommissioning of Biz Ops system codes and we have agreed that it is not essential to verify if a system code is valid or not by making HTTP call to [biz-ops-system-codes](https://system-codes.in.ft.com/) we can simply use a regex patter to check if it is a valid Biz ops system code. 

## What
- Replaces the HTTP call with a simple regex that checks whether a given system code follows a valid Biz ops system code format

### Anything in particular you'd like to highlight to reviewers?
The regex I have used is taken from [Biz ops schema](https://github.com/Financial-Times/biz-ops-schema/blob/9babf227d2cd10bdae33821a2cd10b55c8e95856/schema/string-patterns.yaml#L3) which Biz ops uses to validate system codes

### DIFF
--- a/lib/middleware/v3/parseSystemCodeParameter.js
+++ b/lib/middleware/v3/parseSystemCodeParameter.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const UserError = require('../../utils/usererror');
-const axios = require('axios').default;
 
 module.exports = {
 	parseSystemCodeParameter: parseSystemCodeParameter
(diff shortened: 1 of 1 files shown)


notlee (approved):

notlee:
Thanks @kiyaGu! 

Decommission all the things🔥

origamiserviceuser:
:tada: This PR is included in version v366.4.6 :tada:

PR #629, opened 2024-06-18
chore: add open telemetry to build service

akomiqaia:


notlee (approved):

frshwtr (approved):
LGTM

I am having some confidence issues with the metrics being sent over from Image Service, it is likely that this same config will result in the same for Build Service. However we should be OK to go forward with this and fix-forward once we figure out what's going on.

PR #632, opened 2024-06-20
chore: add dyno name in otel metrics

akomiqaia:


frshwtr (approved):

opentelemetry.js
```diff
@@ -29,6 +29,7 @@ const resource = new Resource({
 	['heroku.app.id']: process.env.HEROKU_APP_ID,
 	['heroku.release.commit']: process.env.HEROKU_BUILD_COMMIT,
 	['heroku.release.creation_timestamp']: process.env.HEROKU_RELEASE_CREATED_AT,
+	['heroku.dyno.name']: process.env.DYNO,
```
sjparkinson, line 32: Hey, what's the thinking behind this resource attribute?

From https://opentelemetry.io/docs/specs/semconv/resource/cloud-provider/heroku/ it's not a standard thing, so I'm curious to understand the use case, and to see if it's something we should get everyone to include.

frshwtr: @sjparkinson for us, the grouping by instance doesn't give us a clear enough picture of what dyno it relates to. Which can mean interpreting cpu/memory charts can be difficult. In prod we typically have two dynos per region, web.1 and web.2. Adding this dyno field helps us see which instance relates to which dyno.

sjparkinson: That makes sense! Thanks for the walk through

PR #634, opened 2024-07-09
Remove graphite metrics

frshwtr:


lib/middleware/sanitize-errors.js
```diff
@@ -8,14 +8,11 @@ module.exports.formatErrorMessage = formatErrorMessage;
...
 function sanitizeErrors() {
 	return (error, request, response, next) => {
-		const {metrics} = request.app.ft;
-
 		// Format the error message if required
 		error.message = module.exports.formatErrorMessage(error);
 
 		// Compilation errors use the non-standard 560 status
 		if (error instanceof CompileError || error.type === 'CompileError') {
-			metrics.count('servererrors.compile', 1);
```
frshwtr, line 18: All of these metrics can be tracked by monitoring the status code for the given endpoint. They are no longer necessary at this level.

lib/middleware/archive.js
```diff
@@ -37,22 +36,18 @@ module.exports = function (app) {
 			}
 
 			if(s3Result.response && s3Result.response.status === 404) {
-				metrics.count('archive.404', 1);
```
frshwtr, line 40: Equally these can also be tracked with auto instrumentation.

notlee: How?

frshwtr: > How?

As the error code is sent through to be the response of this request, auto instrumentation will report on. We can create dashboard views that look at this specific endpoint and count the responses with the 400/200/500 codes etc.

notlee: 👍 Somehow I wasn't thinking that this is all on the one endpoint.

notlee (approved):

